PR #1, opened 2019-06-26
Initial test cases

MatthewHambley:
Obviously this is only a start but I present a number of simple test cases which Fab will need to cope with.


matthewrmshin:
I guess we can also try to migrate relevant tests from `fcm make` to this project as well.

PR #2, opened 2019-06-27
Start design

MatthewHambley:
Started some design documentation with UML diagrams for a graph representation. This is in no way complete work or necessarily representative of any final design.

The files may be converted into graphics using the PlantUML tool.

### DIFF
--- a/Documentation/Design/GraphClass.puml
+++ b/Documentation/Design/GraphClass.puml
@@ -1,0 +1,128 @@
+@startuml
+
+skinparam class {
+    BackgroundColor LemonChiffon
+    BackgroundColor<<hypothetical>> LightSkyBlue
+}
+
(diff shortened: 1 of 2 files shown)


stevewardle:
Perhaps we should include the generated graphics alongside?  (or is the plan to put these directly onto the wiki?)

MatthewHambley:
I have a strong aversion to manually generating renders and uploading them to the wiki. And generated anything simply doesn't belong in a VCS repository.

There seems to be a mechanism to have the wiki make use of a render server: https://stackoverflow.com/questions/32203610/how-to-integrate-uml-diagrams-into-gitlab-or-github

stevewardle:
Yeah in hindsight I agree with you.  Are you wanting this merged at this point?

MatthewHambley:
Lets get it on and we can try generating the images later.

PR #5, opened 2019-07-02
Add an example of using Luigi to control workflow

stevewardle:
Some example code which came out of experimenting with Luigi to asses whether it might be suitable.  I've added it within an "Experimental" folder whose contents we can consider volatile/expendable but might be useful for storing prototyping type work

### DIFF
--- a/Experimental/DAG/luigi_test.py
+++ b/Experimental/DAG/luigi_test.py
@@ -117,9 +117,3 @@ if __name__ == "__main__":
     luigi.build(
         build_manifest,
         local_scheduler=True)
-
-
-        
-
-
-
--- a/Experimental/DAG/luigi_test.py
+++ b/Experimental/DAG/luigi_test.py
@@ -1,0 +1,125 @@
+import os
+import re
+import subprocess
+from glob import glob
+
+import luigi
+
+# Let's define the relationships between the source files here. In practice
+# this information would come from Fab, perhaps it would be stored in
+# something like a sqlite database or similar
+DEPENDENCIES = {
+    "test_exec":
+        ["test_main.f90",],
+    "test_main.f90":
+        ["test.f90",],
+    }
+
+# Similarly another database that gives us information about how to compile
+# each given source. This would likely be more complicated in a final version
+# of Fab; the compiler may be represented as a class or something else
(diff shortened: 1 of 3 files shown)


Experimental/DAG/luigi_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+# Let's define the relationships between the source files here. In practice
+# this information would come from Fab, perhaps it would be stored in
+# something like a sqlite database or similar
+DEPENDENCIES = {
+    "test_exec":
+        ["test_main.f90",],
+    "test_main.f90":
+        ["test.f90",],
+    }
```
MatthewHambley: In an actual build `test_exec` would depend on `test_main.o`, not the source file. And `test_main.o` would depend on `test_main.f90` and `test.mod`.

Is that relevant to this example?

Experimental/DAG/luigi_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+# Similarly another database that gives us information about how to compile
+# each given source. This would likely be more complicated in a final version
+# of Fab; the compiler may be represented as a class or something else
+COMPILERS = {
+    "test_exec": ["ifort",],
+    "test_main.f90": ["ifort", "-c" ],
+    "test.f90": ["ifort", "-c"],    
+    }
```
MatthewHambley: This highlights the issue noted above. `test_exec` is built by `ifort` but `test_main.f90` is not. Rather it is the input **to** `ifort`.

Experimental/DAG/luigi_test.py
stevewardle: You're right.  I did it this way to keep it simpler with less steps to worry about (and to be honest it was already becoming cumbersome even ignoring those extra layers of relationships between things)

Experimental/DAG/luigi_test.py
stevewardle: Agreed - I expect if this were ever progressed towards a more complete example the `COMPILERS` dictionary wouldn't be shared by both the compilation and linking tasks, which would make the inconsistency there less apparent

MatthewHambley:
I am happy with including this example on trunk.

PR #6, opened 2019-07-10
Fix Fortran submodule test case

matthewrmshin:
A function interface had a typo. The fix ensures that the function result is defined correctly. Noticed this while working on metomi/fcm#256.

### DIFF
--- a/TestCases/FortranSubmodule/class_mod.f90
+++ b/TestCases/FortranSubmodule/class_mod.f90
@@ -20,7 +20,7 @@ module class_mod
     function how_much_if(this)
       import foo_type
       class(foo_type), intent(inout) :: this
-      integer :: how_much
+      integer :: how_much_if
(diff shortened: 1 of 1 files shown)


matthewrmshin:
An `implicit none` is already at the top of the source file. Perhaps the check does not apply to a function type?

MatthewHambley:
Fortran scoping is a little unorthodox. Obviously it defined the orthodoxy back in the day but things have moved on since then.

The "implicit none" in the module header applies only to the module. It does not apply to the program units within the module. So all procedures, including interface declarations, need their own. It's a frustration and leads to people forgetting them. Case in point. It also adds to the boiler-plate city nature of modern Fortran source code.

matthewrmshin:
I should now have completed the missing `implicit none` statements in the changed source file. I'll leave it to you to do the others, if necessary.

PR #7, opened 2019-07-10
Missing implicit

MatthewHambley:
It looks like I goofed when approving #6. Some of the "implicit none"s were put in the wrong place. Obviously neither myself nor the developer actually tried typing `make` at any point.

This is now fixed and a missing "implicit" statement is added.

I've also removed a couple of generated source files which shouldn't have made it into the repository.


matthewrmshin:
On *clock skew*, it normally means that your desktop and the storage device (e.g. via NFS) have their clocks out of sync. (And that's also why a timestamp based build system is problematic!)

MatthewHambley:
Apparently renaming the branch to cover this wider remit caused the request to be closed. Work will continue on #8.

PR #8, opened 2019-07-11
Test case tidy

MatthewHambley:
These changes tidy up the test cases in a number of ways. They were originally #7 but become wider than the initial scope of that request. Apparently renaming the branch kills the pull request.

Major changes on this branch are:

 - Removed some generated source which shouldn't have been under version control anyway
 - Fixed misplaced and missing `implicit none` statements
 - Removed the use of `ALWAYS` targets as they don't add anything
 - Added information about required external tools to relevant Make files and added traps for them as well.
 - Tweaked the submodule test to better describe where the numbers come from
 - All tests may be run from the parent directory
 - Added run and verification to each test

This last point has complicated the Make files a bit and may be overkill for what the reviewer originally wanted.


MatthewHambley:
There's a slight wrinkle regarding default values for the compiler macros (FC, CC) which is that make has opinions on what they should be. Opinions from the 70s. So what should be a simple matter becomes a little more complicated.

Still those tests which can be run with any compiler should now support using whatever your FC or CC macros are set to. Some tests require a particular compiler due to specific command line arguments being used. These have been commented.

I also fixed the "clean" targets to not try and delete files which are not created. As for the pFUnit and PSyclone test cases, they do not create executables by design. They are supposed to be minimal test cases after all. They provide just enough to exercise the behaviour we are interested in. The "SimpleLFRic" test combines everything into a full example.

PR #9, opened 2020-01-13
Created workflow for testing on push

MatthewHambley:
This is the start of our continuous integration system.


MatthewHambley:
> Looking at the output from the tests it seems like they are failing because we don't yet have a `requirements.txt` - I guess this is because we don't currently have any requirements, but perhaps we can avoid the error by having an empty copy of the file? Otherwise looks fine to me

The change wasn't ready for review but it is now.

PR #10, opened 2020-01-13
Skeleton production project with packaging

MatthewHambley:
This should build a package which can be uploaded directly to PyPI. That is untested and for another ticket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,7 @@ features which do not target scientific software development.
 ## Licence
 
 The software is made available under a 3-clause BSD licence.
+
+## Installation
+
+You may simply use `pip install foo`.
--- a/README.md
+++ b/README.md
@@ -15,4 +15,4 @@ The software is made available under a 3-clause BSD licence.
 
 ## Installation
 
-You may simply use `pip install foo`.
+You may simply use `pip install model-fab`.
--- a/README.md
+++ b/README.md
@@ -15,4 +15,4 @@ The software is made available under a 3-clause BSD licence.
 
 ## Installation
 
-You may simply use `pip install model-fab`.
+You may simply use `pip install sci-fab`.
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,26 @@
+import setuptools
+
+with open('README.md', 'r') as handle:
+    long_description = handle.read()
+
+setuptools.setup(
+    name='fabricate',
+    version='0.1.dev0',
+    author='',
+    author_email='',
+    description='Build system for scientific software',
+    long_description=long_description,
+    long_description_content_type='text/markdown',
+    url='https://github.com/Metomi/fab',
+    packages=setuptools.find_packages(),
+    classifiers=[
+        'Development Status :: 1 - Planning',
+        'Environment :: Console',
+        'License :: OSI Approved :: BSD License',
+        'Operating System :: OS Independent',
+        'ProgrammingLanguage :: Python :: 3',
+        'Topic :: Scientific/Engineering',
+        'Topic :: Software Development :: Build Tools'
+        ],
+    python_requires='>=3.6'
+)
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,7 @@
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+
 import os.path
 import setuptools
 
--- a/setup.py
+++ b/setup.py
@@ -1,18 +1,32 @@
+import os.path
 import setuptools
 
-with open('README.md', 'r') as handle:
-    long_description = handle.read()
+_here = os.path.dirname(__file__)
+
+with open(os.path.join(_here, 'README.md'), 'r', encoding='utf-8') as handle:
+    _long_description = handle.read()
+
+with open(os.path.join(_here, 'source', 'fab', '__init__.py'),
+          encoding='utf-8') as handle:
+    for line in handle:
+        bits = line.split('=', 1)
+        if bits[0].strip().lower() == '__version__':
+            _version = bits[1].strip()
+            break
+    else:
+        raise RuntimeError('Cannot determine package version.')
 
 setuptools.setup(
-    name='fabricate',
-    version='0.1.dev0',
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -0,0 +1,45 @@
...
+    for line in handle:
+        bits = line.split('=', 1)
+        if bits[0].strip().lower() == '__version__':
+            _version = bits[1].strip()
+            break
+    else:
+        raise RuntimeError('Cannot determine package version.')
+
+setuptools.setup(
+    name='model-fab',
```
stevewardle: Now that I see it written below I think I actually prefer `sci-fab`

setup.py
```diff
@@ -0,0 +1,45 @@
...
+    long_description=_long_description,
+    long_description_content_type='text/markdown',
+    url='https://github.com/Metomi/fab',
+    package_dir={'': 'source'},
+    packages=setuptools.find_packages(where='source'),
+    classifiers=[
+        'Development Status :: 1 - Planning',
+        'Environment :: Console',
+        'License :: OSI Approved :: BSD License',
+        'Operating System :: OS Independent',
```
stevewardle: I've got a feeling we may struggle to keep this truly "OS Independent" as time goes on, but we can revise this later I suppose

setup.py
```diff
@@ -0,0 +1,45 @@
...
+        'ProgrammingLanguage :: Python :: 3',
+        'Topic :: Scientific/Engineering',
+        'Topic :: Software Development :: Build Tools'
+        ],
+    python_requires='>=3.6, <4',
+    entry_points={'console_scripts': ['fab=fab.__main__:main']},
+    project_urls={
+        'Bug reports': 'https://github.com/Metomi/fab/issues',
```
stevewardle: Lowercase `metomi`? (and on the line below too)

source/fab/__init__.py
```diff
@@ -0,0 +1,5 @@
+'''
+Flexible build system for scientific software.
+'''
+
+__version__ = '0.1.dev0'
```
stevewardle: Just a passing thought - I think we will eventually want to go with a YYYY.MM.D style of versioning (though I doubt it matters in the build up to the first release version)

source/fab/__main__.py
```diff
@@ -0,0 +1,38 @@
...
+def parse_cli() -> argparse.Namespace:
+    '''
+    Parse the command line for arguments.
+    '''
+    description='Flexible build system for scientific software.'
+    parser = argparse.ArgumentParser(add_help=False,
+                                     description=description)
+    parser.add_argument('-help', '-h', '--help', action='help',
+                        help='Print this help and exit')
+    parser.add_argument('-version', action='version', version=fab.__version__)
```
stevewardle: I think convention is generally to have `--` as a prefix to any flags that aren't single character?  (and on the `-verbose` option below as well)

README.md
```diff
@@ -12,3 +12,7 @@ features which do not target scientific software development.
 ## Licence
 
 The software is made available under a 3-clause BSD licence.
+
+## Installation
+
+You may simply use `pip install model-fab`.
```
stevewardle: As per my other comment further down - maybe `sci-fab` is better?

source/fab/__init__.py
MatthewHambley: Not that I'm against changing it right now, but why do you think that?

source/fab/__main__.py
MatthewHambley: Only on GNU projects. Other realms prefer different approaches. In the particular case of "help" I want to cover all bases I can think of. It should always be easy to get help.

setup.py
MatthewHambley: Happy to oblige.

setup.py
MatthewHambley: Fair enough. I've changed it to Posix, which includes Linux despite it not being Posix compliant.

setup.py
MatthewHambley: Quite correct.

source/fab/__init__.py
stevewardle: From past experience I find having the major-minor style of versioning means you are forever questioning "should this be a major release?" whereas with the date based one it feels more like it gives every release equal(ly insignificant) weighting and takes that decision out of your hands

source/fab/__main__.py
stevewardle: Fair enough with help, I agree... I guess we need to decide if we are to follow GNU convention or not (just thinking about a lot of the other tools we use like Rose, Cylc etc. all seem to follow it)

source/fab/__init__.py
MatthewHambley: I tend to think the question of whether it is major (i.e. breaks backward compatibility) or not is an important one to ask on every release. However not strongly enough to argue the point. I have changed it and made a guess at when our first release will be.

source/fab/__main__.py
MatthewHambley: In the case of Rose and Cylc, they sort-of follow the GNU standard. The glibc documentation has this to say: https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

As you can see, the GNU "long options" extension specifies `--<name>=<value>` and while Rose and Cylc does support that it also supports `--<name> <value>`.

Sun were proponents of the long option with single hyphen which is reflected in Java. I tend to prefer that mode but it does mean that you can no longer use single letter options. After all does `-cat` mean "cat" or does it mean "c", "a" and "t"?

But there isn't even agreement within Posix tools, look at `find` which offers options like `-name`.

Basically we just have to choose what we like and stick to it.

stevewardle: Okay well I think we have one vote for each, @GiorgiaLine for the deciding vote?

source/fab/__init__.py
stevewardle: Agreed

source/fab/__main__.py
GiorgiaLine: I think I prefer the GNU convention for this, if only because I like being able to use single letter options.

MatthewHambley: As you wish.

PR #14, opened 2020-01-16
Add very basic Fortran source sanitisation

stevewardle:
This is really just to give us some very basic functionality that we can use as a base to think about putting in the various testing workflows that we want to have.  It adds a method for going through the source of a (Fortran) file but with a bit of cleaning to try and give the subsequent parsing steps we'll want to do an easier time.

I'm unsure about what kind of structure we will eventually want in terms of language-specific stuff so for now I've assumed we might compartmentalise things via `fab.language.<language_name>` but in time I expect this to change.  Currently `fab` just spits out the sanitised source in full, so at least you can tell it's doing something!

### DIFF
--- a/source/fab/__main__.py
+++ b/source/fab/__main__.py
@@ -9,8 +9,15 @@ Command-line interface to Fab build tool.
 import argparse
 import logging
 import sys
+import os
 
 import fab
+from fab.language.fortran import reader
+
+_extensions = [
+    'F90',
+    'f90',
+    ]
 
 
 def parse_cli() -> argparse.Namespace:
@@ -27,7 +34,19 @@ def parse_cli() -> argparse.Namespace:
                         help='Print version identifier and exit')
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Produce a running commentary on progress')
+    parser.add_argument('source',
+                        help='The path to the source tree to build')
     return parser.parse_args()
+
+
+def sourcepath_iter(rootpath: str) -> str:
+    '''
+    Return files we can process from the source tree.
+    '''
+    for root, _, files in os.walk(rootpath):
+        for name in files:
+            if name.split(".")[-1] in _extensions:
+                yield os.path.join(root, name)
 
 
 def main() -> None:
@@ -43,3 +62,8 @@ def main() -> None:
         logger.setLevel(logging.INFO)
     else:
         logger.setLevel(logging.WARNING)
+
+    for sourcefile in sourcepath_iter(arguments.source):
+        print(f'Processing {sourcefile}:')
+        print(''.join(reader.sourcefile_iter(sourcefile)))
+        print(f'Processed {sourcefile}\n')
--- a/source/fab/__main__.py
+++ b/source/fab/__main__.py
@@ -36,7 +36,7 @@ def parse_cli() -> argparse.Namespace:
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Produce a running commentary on progress')
     parser.add_argument('source',
-                        help='The path to the source tree to build')
+                        help='The path of the source tree to build')
     return parser.parse_args()
 
 
--- a/source/fab/__main__.py
+++ b/source/fab/__main__.py
@@ -9,14 +9,15 @@ Command-line interface to Fab build tool.
 import argparse
 import logging
 import sys
-import os
+from pathlib import Path
+from typing import Generator
 
 import fab
 from fab.language.fortran import reader
 
 _extensions = [
-    'F90',
-    'f90',
+    '.F90',
+    '.f90',
     ]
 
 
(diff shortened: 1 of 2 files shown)


MatthewHambley:
> I'm unsure about what kind of structure we will eventually want in terms of language-specific stuff so for now I've assumed we might compartmentalise things via `fab.language.<language_name>` but in time I expect this to change. Currently `fab` just spits out the sanitised source in full, so at least you can tell it's doing something!

From my point of view I want something I can use to set up the system testing. Ultimately we will use Fab on our various test cases and check the compiled output with a known-good-output. As such it would probably be best for me if the output was a concatenated file of all the source seen.

I don't know how this will work for anyone else.

stevewardle:
> From my point of view I want something I can use to set up the system testing. Ultimately we will  use Fab on our various test cases and check the compiled output with a known-good-output. As such it would probably be best for me if the output was a concatenated file of all the source seen.

If that will be suitable then I think you can already do:

`fab TestCases/ > output.txt`

and get something pretty close to what you're describing.  With a few details that could maybe be tweaked:

- It includes the `Processing <file>` and `Processed <file>` messages, which could maybe be tweaked (suggest just including the first of the two and maybe giving it more of a banner appearance to make the output easier to parse visually?)

- Empty lines (e.g. where the entire line contents have been sanitised) end up making the output look a bit sparse.  I suppose I could have it only output lines with actual content on them? (this would also mean later steps also don't have to step through all the empty lines)

stevewardle:
I've also noticed in the CI testing that since one of the 3 versions we test is `3.5` and the literal string interpolation ("f strings") wasn't added until `3.6` it means we're going to have to choose between using them and drop the `3.5` testing, or stick to `.format` for strings and support all of the current versions.  Any thoughts?

MatthewHambley:
> If that will be suitable then I think you can already do:
> 
> `fab TestCases/ > output.txt`

I think that would probably be sufficient. If there are simple changes you can make to improve the comparability of such files then so much the better.

We might want to do something like `fab TestCases/PSyClone | md5sum -` and avoid having to keep copies of the expected output lying around. On the other hand we might prefer to be able to get a diff between expected output and actual output.

MatthewHambley:
> I've also noticed in the CI testing that since one of the 3 versions we test is `3.5` and the literal string interpolation ("f strings") wasn't added until `3.6` it means we're going to have to choose between using them and drop the `3.5` testing, or stick to `.format` for strings and support all of the current versions. Any thoughts?

There's nothing magical about the versions chosen, I just copied it from the Stylist repository. If we want to use a feature that is not available from Python 3.5 then we can just drop that version from the testing and packaging.

"F strings" seem like a nice feature and even when it's automatic there's no inherent virtue in testing against too many Python versions. So I would support dropping 3.5 support. Raise an issue and assign it to me if you are still pro-f-string.

stevewardle:
Okay for now I've not used the "f strings" feature so that this PR can go through without failing the CI but I'll raise an issue (I think we should go with f strings, it's said to be the most efficient string formatter)

I've tidied up the output a little as discussed above - in the end I figured it might as well condense all white-space down as much as possible to give whatever processes or hashes it next minimal work

stevewardle:
Unless anyone has any strong objections I think we may as well review/commit this - as mentioned above I'm not entirely happy with the structure it imposes but I think once the project has even this tiny amount of meat on the bones it will give us all a point to start working from on different things

source/fab/language/fortran/reader.py
```diff
@@ -0,0 +1,43 @@
...
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+
+'''
+Source reading tools for Fortran format.
+'''
+import re
+
+
+def sourcefile_iter(filename: str) -> str:
```
MatthewHambley: As above, thus function yields a generator of strings, not strings themselves.

Are you aware of `pathlib` which has been added to Python 3 and which improves handling of file paths? Passing `filename` as a string is fine for the moment but it's worth looking into for the future.

source/fab/__main__.py
```diff
@@ -27,9 +34,21 @@ def parse_cli() -> argparse.Namespace:
                         help='Print version identifier and exit')
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Produce a running commentary on progress')
+    parser.add_argument('source',
+                        help='The path to the source tree to build')
     return parser.parse_args()
 
 
+def sourcepath_iter(rootpath: str) -> str:
```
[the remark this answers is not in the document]
stevewardle: Ah okay, thanks - I did wonder whether this was right (I've not made a lot of use of annotations before)

source/fab/language/fortran/reader.py
stevewardle: I'll see what I can do (it'd be good to avoid putting off using something more suitable especially when it's such a small set of code to update)

source/fab/__main__.py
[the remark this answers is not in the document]
stevewardle: Should be fixed now - although it ends up being `Generator[Path, None, None]` as per the other comment I've change it to use `pathlib` objects for the paths

source/fab/language/fortran/reader.py
stevewardle: Hopefully that seems okay?  `pathlib` is certainly a lot neater

stevewardle:
Note that the checks here will work correctly once #19 gets merged and this is rebased

stevewardle:
It worked :)  So now I think this is ready for another approval and to be merged

Issue #15, opened 2020-01-17
Drop testing of Python 3.5


stevewardle:
As noted in #14  - our current set of tests includes one which uses Python version `3.5`, but since we'd ideally like to make use of the new "f-strings" feature (added at `3.6`) from the outset, let's update the testing to only use version `3.6` onwards

stevewardle:
Closed by #16

PR #16, opened 2020-01-17
Dropped Python 3.5 from testing and packaging.

MatthewHambley:
As per issue #15, this change means we no longer test or package for Python 3.5.

Awkwardly this request is run under the existing regime so runs all 3. If you check the original branch, green tick above the file listing, you will see that it ran only 3.6 and 3.7.


.github/workflows/pythonpackage.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.5, 3.6, 3.7]
+        python-version: [3.6, 3.7]
```
stevewardle: Can we test 3.8 as well?

stevewardle:
Also note - the checks should run correctly on `metomi/fab` once #19 gets committed and this is rebased

stevewardle:
Well the fix to the testing worked, although it looks like we have a slight issue in that since this test removes testing of 3.5 the checks won't fully complete - I think once this is through review and (essentially) ready to merge we'll just de-activate the 3.5 test requirement in the settings so that we can do the commit (it'll need to be de-activated afterward anyway)

.github/workflows/pythonpackage.yml
MatthewHambley: Your wish is my command.

PR #19, opened 2020-01-20
Add reference to readme about command line interface and fix the workflow

stevewardle:
This is mostly being done to confirm a theory about the reason why our actions aren't running on `metomi/fab` - namely that since we modified the repository settings _during_ the other PRs which were raised that has somehow put them into a broken state.  Since this PR is being created _after_ the changes perhaps it will work correclty

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -1,6 +1,6 @@
 name: Python package
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


stevewardle:
Aha!  I've worked out what the problem was.  Our file controlling the workflor was set to activate on `push` - and we can see it working on our own forks (when pushing changes to them) and also on `metomi` when things are _merged_.  However raising a Pull Request against `metomi` does not count as a push, so we need to change the workflow to activate on either `push` or `pull_request`.  It looks like it's working to me!

stevewardle:
Ready for review now if somebody would like to take a look - up to you whether we keep the part I added to the `README` or not (I just needed to add _something_ to test a new PR)

PR #20, opened 2020-02-03
Make sure Flake is actually run and fix problems it finds

MatthewHambley:
I noticed that although Flake was running it was ignoring all errors. Hardly worth the effort!

This change allows it to fail and fixes all the problems it found.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -21,13 +21,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -e .
-    - name: Lint with flake8
+    - name: Code style check with flake8
       run: |
         pip install -e .[dev]
-        # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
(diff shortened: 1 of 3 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -21,13 +21,10 @@ jobs:
...
         pip install -e .
-    - name: Lint with flake8
+    - name: Code style check with flake8
       run: |
         pip install -e .[dev]
-        # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+        flake8 . --count --show-source --statistics
```
stevewardle: Can you elaborate a little on what the difference is following the change?  What about the flags before was causing it to not check anything?

MatthewHambley: The `--select=...` argument causes flake to only check the specified rules. Meanwhile `--exit-zero` means that even errors do not cause flake to exit with an error. Of course there were two calls to flake with different argument lists. This is replaced with a single invocation which tests for everything and will exit with an error if any errors are found.

stevewardle: Sounds good to me - thanks

Issue #13, opened 2020-01-16
System testing


MatthewHambley:
Once we have a tool which can do something with a source tree we should see how we can include system testing in our workflow.

We can use the trees held in TestCases.

MatthewHambley:
I'm closing this ticket as the associated pull request has been merged.

PR #21, opened 2020-02-04
Prototype system testing

MatthewHambley:
For your consideration: System testing using one of the test cases we harvested during requirements capture.

This makes use of the `systest` framework. If we choose to adopt it we might well want to raise some pull-requests against it to migrate some of the stuff I have added upstream.

This work is in reference to #13.


MatthewHambley:
I think I have discovered an issue with the workflow so let me check that before making any decisions as to whether to go ahead with this change.

stevewardle:
For now I think this approach using `systest` looks good.  As you say we can try to engage with the author and see how we go - worst case if we run into problems we can always change it later.  @GiorgiaLine - what do you think?  (and perhaps if you'd like to do the review as well?)

MatthewHambley:
Panic over, it looks like everything is working as expected.

MatthewHambley:
> Not sure if it matters since it's a text file, but I beleve there's an open string and 2 missing close-parenthesis on line 7 of system-tests/MinimalFortran/expected.txt?

What you have detected there is a short-cut taken by Steve's dependency analyser code. It's cleaved the line at an exclamation mark, assuming it to be a comment. In this case it's actually within a string so strictly speaking should not be touched.

So the expected text is the output of the program as it stands. Clearly that is not right but that is a problem with the program, not with the testing.

stevewardle:
> What you have detected there is a short-cut taken by Steve's dependency analyser code. It's cleaved the line at an exclamation mark, assuming it to be a comment. In this case it's actually within a string so strictly speaking should not be touched.

I was aware that this would be a case that it wouldn't quite process right - but I think I convinced myself that since the next step would be to look for program units and imports, and these are always identified by things that related to the _beginning_ of a line, that it wasn't a big deal.  I might be wrong about that of course!  In which case we'll need to fix it

Issue #28, opened 2020-02-07
Make sure a suitable set of tools/methods exist to query the tables in the database


stevewardle:
As long as the database is maintained using the SQLite DBMS it will always be possible to use the command-line tool associated with that library to interrogate the database. This ticket is concerned with providing a simple tool to allow developers not familiar with SQL to explore the data contained therein.
Views which should be available through the tool are:
  * A list of files and attendant information
    * Hash
    * Contained program units
  * Program unit dependency

While it would nice for this to be presented through a GUI a well implemented terminal interface would suffice.
It is also concerned with the APIs associated with database access. It is not unreasonable for some levels of the API to expose SQL but there should also be higher level methods which do not.

matthewrmshin:
(There are plenty of off-the-shelves SQLite browsing tools available. Search for `sqlite browser`.)

MatthewHambley:
> (There are plenty of off-the-shelves SQLite browsing tools available. Search for `sqlite browser`.)

That's a good suggestion but I suspect we are looking for something more than simply inspecting the content of the tables. I think we need something which embodies pre-written queries as well. What do others think?

stevewardle:
That's kind of what I was thinking... like I can picture a scenario where Fab is doing something unexpected - maybe it's falling over because something it thinks is compiled actually hasn't been - and so I want to check quickly whether the database contains the dependency that should have caused the missing thing to be compiled... Being able to dump out the contents in various ways would be helpful for this kind of thing (I'm not sure if that's what an off-the-shelf tool gives you or not)

matthewrmshin:
Make sense. There are multiple ways to do this. You can write SQL files to do this or Python utilities to do this.

MatthewHambley:
With the completion of #39 and #57 I consider this initial ticket complete. There will be further work as additional features are added.

PR #39, opened 2020-03-18
New "fab-dump" tool to view database content

MatthewHambley:
The addition is a simple tool which simply dumps database content to the terminal. At the moment it's output looks a lot like the main program's output but that will change as the main program starts to actually do something.

Implicit in this change is the addition of support for multiple entry points. i.e. multiple executables.

This is part of #28.


stevewardle:
Back to you to update to master

source/fab/application.py
```diff
@@ -56,3 +57,32 @@ def run(self, source: Path):
...
+
+    def run(self, stream=sys.stdout):
+        file_view = FileInfoDatabase(self._state)
+        print("File View", file=stream)
+        for filename in file_view.get_all_filenames():
+            file_info = file_view.get_file_info(filename)
+            print(f"  File   : {file_info.filename}", file=stream)
+            # Where files are generated in the working directory
+            # by third party tools, we cannot guarantee the hashes
+            if file_info.filename.match(f'{self._workspace}/*'):
```
stevewardle: Noting the duplication here with what the main application is doing (albeit in a temporary way) - could we not have the main application produce its output via the `Dump` class?  I can see arguments for both - maybe not tying the two together like that is for the best (although again, it's only temporary)

source/fab/entry.py
```diff
@@ -0,0 +1,105 @@
...
+import logging
+from pathlib import Path
+import sys
+
+import fab.application
+
+
+def fab_cli() -> argparse.Namespace:
+    """
+    Parses command line arguments for the core "fab" tool.
```
stevewardle: The way we refer to the tool here as `the core "fab" tool` is inconsistent with the way we refer to it on L44 below, as `the core Fab build tool` - it doesn't matter much but could pick one and stick to it

source/fab/entry.py
```diff
@@ -0,0 +1,105 @@
...
+
+    arguments = dump_cli()
+
+    if arguments.verbose:
+        logger.setLevel(logging.INFO)
+    else:
+        logger.setLevel(logging.WARNING)
+
+    if not arguments.workspace:
+        arguments.workspace = Path.cwd() / 'working'
```
stevewardle: Should we be bothered that the location of the default working directory is subtly different between the two tools?  I understand that the dumper tool cannot use the source directory (as there's no such concept) but perhaps the default location for the main tool should be in `cwd` also rather than in `source`?  This also ties into the fact that perhaps we don't want to be adding anything to the source directory if we can help it.  

Could do this as a followup though as it's not really in scope

system-tests/FortranDependencies/first.f90
```diff
@@ -2,7 +2,7 @@
 ! For further details please refer to the file COPYRIGHT
 ! which you should have received as part of this distribution
 !
-program first
+program first_mod
```
stevewardle: Why did you rename this to `first_mod`?  I guess to highlight that the unit name needn't match the source filename?  Which is fine, I just wonder about using `_mod` - which is conventionally used for modules

system-tests/test.py
```diff
@@ -13,51 +13,48 @@
...
-        environment = {
-            'PYTHONPATH': 'source',
-            'PATH': os.environ.get("PATH", '')
-            }
+        script = "import sys; import fab.entry; " \
+                 f"sys.exit(fab.entry.{self._entry_point}())"
+        command = ['python3', '-c', script,
+                   '-w', self._working_dir]
+        command.extend(self._arguments)
+        new_path = os.environ.get('PATH') \
```
stevewardle: This has lost the protection against the (extremely unlikely) case that `PATH` isn't set - it's probably fine but just pointing it out

system-tests/test.py
```diff
@@ -145,12 +186,21 @@ def _assert_diff(self, first, second):
...
+            DumpTestCase(root_dir / 'MinimalFortran')
+        ],
+        [
+            FabTestCase(root_dir / 'FortranDependencies'),
+            DumpTestCase(root_dir / 'FortranDependencies')
+        ],
+        [
+            FabTestCase(root_dir / 'FortranPreProcess',
+                        fpp_flags='-DSHOULD_I_STAY=yes'),
+            DumpTestCase(root_dir / 'FortranPreProcess')
```
stevewardle: So just checking - these are executed in order?  So the `DumpTestCase` will be running on the pre-processed result?  (that seems to be the case based on the expected output files)

source/fab/application.py
MatthewHambley: I think it best to keep them separate. The dumper behaviour is the desired behaviour for that tool. It absolutely isn't the desired behaviour for the build tool. I don't think we gain anything by conflating them other than drawing a false equivalence.

source/fab/entry.py
MatthewHambley: I take your points about divergant behaviour and not putting the working directory in the source directory. I'm not sure what the correct default behaviour should be. I think postponing this to a different ticket is the way to go. See #43.

system-tests/FortranDependencies/first.f90
MatthewHambley: Crikey! That was a mistake. `first` is a program so should never have the `_mod` suffix. Having program units which don't match their filenames is a good call though. I've renamed "second" to "two".

system-tests/test.py
MatthewHambley: I've replaced this with a more comprehensive solution which should solve that problem. Unlikely as it is.

system-tests/test.py
MatthewHambley: That is correct. I have commented the structure to make it clearer.

source/fab/application.py
stevewardle: Yeah I agree - just thought I'd mention it to show I was paying attention!

system-tests/FortranDependencies/first.f90
stevewardle: Easily done!

PR #40, opened 2020-03-18
Test hashing algorithms for speed

MatthewHambley:
There was some concern that the Adler32 checksum used in #38 was a pure python implementation and thus too slow. This didn't seem likely as comes from the `zlib` package and therefore, presumably, uses the `zlib` implementation.

A simple experiment was written to benchmark the various hashing algorithms. This confirms what we expected, specifically that the two data integrity checksums outperform the cryptographic hashes by 2 orders of magnitude.


Experimental/BenchmarkHashes/hashbench.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
stevewardle: Perhaps we want a copyright header here (I note that I didn't put one in the existing `Experimental` test code, but I probably should have (not sure if we can justify adding them as part of this change?  Slightly out of scope but raising a PR just to modify them feels like hassle

Experimental/BenchmarkHashes/psykal_lite_mod.F90
```diff
@@ -0,0 +1,2669 @@
+!-----------------------------------------------------------------------------
```
stevewardle: I know this comes from another project but I think we should modify this copyright header in some way, or add a clarifying line at the end to make it clear that this is not applicable to the Fab project and is just from this example file's original implementation

Experimental/BenchmarkHashes/hashbench.py
MatthewHambley: Done

Experimental/BenchmarkHashes/psykal_lite_mod.F90
MatthewHambley: And done.
I've copied the licence text from the donor project, as I should have done in the first place.

PR #41, opened 2020-03-20
Put initial classes in place for running tasks and implement Fortran PreProcessor

### LINKED ISSUE #32: Implement Fortran preprocess transformation

stevewardle:
Building on the framework put in place by the `Analyser` this aims to implement a more general `Task` class, of which an `Analyser` becomes a member, but also other types of task that Fab needs to perform.  As an initial example it implements a type of task which runs a command, and this is used to run a simple preprocessing step

Closes #32


stevewardle:
Over to you @MatthewHambley - I'm expecting this might need a bit of work as I've tried to fit things in around what was there as best as I can, but I'm a little unsure about some aspects

setup.py
```diff
@@ -50,6 +50,6 @@
     extras_require={
         'dev': ['flake8', 'mypy'],
         'unit-test': ['pytest', 'pytest-cov'],
-        'system-test': ['systest']
+        'system-test': ['systest', ]
```
MatthewHambley: What is the reason for the addition of a spurious comma? Is it something which your editor insists on?

source/fab/__main__.py
```diff
@@ -29,7 +29,9 @@ def parse_cli() -> argparse.Namespace:
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Produce a running commentary on progress')
     parser.add_argument('-w', '--workspace', metavar='FILENAME', type=Path,
-                        help='Directory for working files.')
+                        help='Directory for working files')
+    parser.add_argument('--fpp-flags', action='store', type=str, default='',
+                        help='Provide flags for Fortran PreProcessor ')
```
MatthewHambley: Using an argument in lue of an actual configuration system is fine but can you add a "TODO" to note that it is a temporary fix.

source/fab/application.py
```diff
@@ -4,36 +4,50 @@
...
 
 class Fab(object):
-    _extensions: Dict[str, Type[Analyser]] = {
-        '.F90': FortranAnalyser,
-        '.f90': FortranAnalyser
+    _extension_map: Dict[str, Union[Type[Task], Type[Command]]] = {
+        '.f90': FortranAnalyser,
+        '.F90': FortranPreProcessor,
     }
+    _command_flags_map: Dict[Type[Command], List[str]] = {}
```
MatthewHambley: Havin `_command_flag` as a class variable and adding to it in the constructor means it will collect mappings from every instance. Is that the behaviour you want?

source/fab/application.py
```diff
@@ -4,36 +4,50 @@
...
         for file in file_db.get_all_filenames():
             info = file_db.get_file_info(file)
             print(info.filename)
-            print(f'    hash: {info.adler32}')
+            # Where files are generated in the working directory
+            # by third party tools, we cannot guarantee the hashes
+            if str(info.filename).startswith(str(self._workspace)):
+                print('    hash: --hidden-- (generated file)')
+            else:
+                print(f'    hash: {info.adler32}')
```
MatthewHambley: This is not ideal but it is also throw-away code so not a problem.

source/fab/language/fortran.py
```diff
@@ -310,11 +310,10 @@ def analyse(self, source: TextReader) -> None:
                     unit_type: str = unit_match.group(1).lower()
                     unit_name: str = unit_match.group(2).lower()
                     logger.debug('Found %s called "%s"', unit_type, unit_name)
-                    self._state.add_fortran_program_unit(unit_name,
-                                                         source.filename)
+                    self._state.add_fortran_program_unit(
+                        unit_name, self._reader.filename)
```
MatthewHambley: We should start compiling a style guide document where we agree on how we are going to handle, for instance,  continuation argument lists. Both the original (my style) and new (your style) are valid. We just need to pick one.

source/fab/language/fortran.py
```diff
@@ -393,5 +392,20 @@ def analyse(self, source: TextReader) -> None:
...
                                       'name': exp[1],
                                       'found': end_name}
-                        raise AnalysisException(
+                        raise TaskException(
                             end_message.format(**end_values))
+        return []
+
+
+class FortranPreProcessor(Command):
+    stdout = False
```
MatthewHambley: This is set but I do not believe it is used anywhere.

source/fab/source_tree.py
```diff
@@ -22,36 +22,56 @@ def visit(self, candidate: Path):
...
+                 extension_map: Mapping[str, Union[Type[Task], Type[Command]]],
+                 command_flags_map: Mapping[Type[Command], List[str]],
+                 state: SqliteStateDatabase, workspace: Path):
         self._extension_map = extension_map
+        self._command_flags_map = command_flags_map
+        self._state = state
+        self._workspace = workspace
 
-    def visit(self, candidate: Path):
+    def visit(self, candidate: Path) -> List[Path]:
```
MatthewHambley: You've changed the call signature of this implementation but not of the abstract it's overriding.

source/fab/source_tree.py
```diff
@@ -22,36 +22,56 @@ def visit(self, candidate: Path):
...
 
 class TreeDescent(object):
     def __init__(self, root: Path):
         self._root = root
 
     def descend(self, visitor: TreeVisitor):
-        visit = [self._root]
-        while len(visit) > 0:
-            candidate: Path = visit.pop()
+        addresses = [self._root]
```
MatthewHambley: Still doesn't seem like the best name but better than `visit`.

system-tests/test.py
```diff
@@ -47,8 +49,13 @@ def teardown(self):
...
     def run(self):
-        command = ['python3', '-m', 'fab', self._test_directory]
-        environment = {'PYTHONPATH': 'source'}
+        command = ['python3', '-m', 'fab']
+        if self._fpp_flags is not None:
+            command.append('--fpp-flags=' + self._fpp_flags)
+        command.append(self._test_directory)
+
+        environment = os.environ
+        environment.update({'PYTHONPATH': 'source'})
```
MatthewHambley: I don't think the wholesale copying of the users environment is a good way to go. It risks pulling in a lot of complicating factors. Instead please try copying only the required variables across. e.g. PATH. It may be necessary to extend rather than replace PYTHONPATH in the future.

system-tests/test.py
```diff
@@ -138,7 +145,9 @@ def _assert_diff(self, first, second):
 
     sequence = [
         FabTestCase(root_dir / 'MinimalFortran'),
-        FabTestCase(root_dir / 'FortranDependencies')
+        FabTestCase(root_dir / 'FortranDependencies'),
+        FabTestCase(root_dir / 'FortranPreProcess',
+                    fpp_flags='-DSHOULD_I_STAY=yes')
```
MatthewHambley: I agree with the developer that we should probably test the undefined situation as well. This will require additional system testing infrastrucgture so I am happy for it to be postponed to a future ticket provided one is created and linked to this review.

unit-tests/source_tree_test.py
```diff
@@ -61,34 +83,71 @@ def test_extension_visitor(tmp_path: Path):
     (tmp_path / 'dir').mkdir()
     bar_file = tmp_path / 'dir' / 'file.bar'
     bar_file.write_text('Second file in subdirectory')
+    baz_file = tmp_path / 'file.baz'  # Doesn't exist
 
     db = SqliteStateDatabase(tmp_path)
     file_info = FileInfoDatabase(db)
+    clear_tracker()
```
MatthewHambley: Can you use test fixtures to clear the tracker?

unit-tests/source_tree_test.py
```diff
@@ -61,34 +83,71 @@ def test_extension_visitor(tmp_path: Path):
...
+    foo_file.write_text('First file')
+    bar_file = tree_root / 'file.bar'
+    bar_file.write_text('Second file')
+    # This file doesn't exist - it will be created by the
+    # command task when run on the "bar" file
+    baz_file = tree_root / 'file.baz'
+
+    db = SqliteStateDatabase(tmp_path)
+    file_info = FileInfoDatabase(db)
+    clear_tracker()
```
MatthewHambley: See previous question about fixtures.

setup.py
stevewardle: No - it's because whilst trying to understand the issues with `cpp` output I thought I might be able to control installing a specific `gcc` version here, and when that didn't work I failed to tidy up after myself fully!

source/fab/__main__.py
stevewardle: Good point

source/fab/application.py
stevewardle: I think the way the constructor adds to it means that it won't collect _duplicates_ at least, but beyond that I'd not really thought too much about it.  I'll move the initialisation into the constructor for now

source/fab/application.py
stevewardle: Agree it's not great, but after some time I couldn't find a way around it - the slightest difference in `GCC` version can cause differences in the generated output - in this case the version used by Github's Actions was adding 2 extra blank lines at the start of the produced file (and there's no way to control this it seems)... I think this is a nicer solution than trying to do something strange in the class (e.g. trimming out all leading blank lines until we              hit the first executable line of the original file, or something, which I considered doing as well)

source/fab/language/fortran.py
stevewardle: I hadn't even noticed this.  I think in this case I _had_ to do it the way I have because the new longer name would pass 80 characters under the "old" style

source/fab/language/fortran.py
stevewardle: It's used by the `CommandTask` class in its `run` method - if it is true it causes the task to capture the stdout and direct it into the file given by the `output_filename` property

stevewardle: Having said that, `False` is the default in the `Command` base class anyway, so I can take it out (apologies if that's what you were getting at to begin with!)

source/fab/source_tree.py
stevewardle: I'm open to suggestions!  You can `visit` and `address` right?  (though I guess address is often used in networking and such)

system-tests/test.py
stevewardle: Okay, I'll give it a try

system-tests/test.py
stevewardle: I think I will postpone this until after #39 goes on, since that already makes some headway in this regard (and I'd rather not have us stomp all over each other's work)

Opened #42 to deal with this

unit-tests/source_tree_test.py
stevewardle: I think I've now done this using `setup_function`

unit-tests/source_tree_test.py
stevewardle: As above

source/fab/application.py
MatthewHambley: For the future we will probably want to examine the resuling files. The hashes are super useful to us internally but are not something we should be making users care about.

source/fab/language/fortran.py
MatthewHambley: A discussion for another forum.

source/fab/source_tree.py
MatthewHambley: I went to networks when I saw "address". How about something simple like `to_visit` or maybe `agenda`?

stevewardle: Yeah okay, I'll go with `to_visit`

PR #49, opened 2020-04-01
Improve unit test coverage

MatthewHambley:
These changes push our unit test coverage as far as it can reasonably go.

I don't think it makes sense to test our abstract types to ensure they are properly abstract. It is a shortcoming in Python that they might not be but I think we can safely put that off until we have spare time.

A few minor issues have been fixed as they were discovered while developing the tests.

### DIFF
--- a/source/fab/database.py
+++ b/source/fab/database.py
@@ -9,7 +9,7 @@ Working state which is either per-build or persistent between builds.
 from abc import ABC, abstractmethod
 from pathlib import Path
 import sqlite3
-from typing import Dict, Iterator, Sequence, Union
+from typing import Dict, Iterator, Optional, Sequence, Union
 
 
 class WorkingStateException(Exception):
@@ -26,14 +26,16 @@ class FileInfo(object):
             message = f"Cannot compare FileInfo with {other.__class__}"
             raise ValueError(message)
         return (str(other.filename) == str(self.filename)) \
-               and (other.adler32 == self.adler32)
+            and (other.adler32 == self.adler32)
 
 
 class DatabaseRows(Iterator[Dict[str, str]]):
-    def __init__(self, cursor: sqlite3.Cursor):
+    def __init__(self, cursor: Optional[sqlite3.Cursor]):
         self._cursor = cursor
 
     def __next__(self) -> Dict[str, str]:
+        if self._cursor is None:
+            raise StopIteration()
         row = self._cursor.fetchone()
         if row is None:
             raise StopIteration()
--- a/source/fab/database.py
+++ b/source/fab/database.py
@@ -9,7 +9,7 @@ Working state which is either per-build or persistent between builds.
 from abc import ABC, abstractmethod
 from pathlib import Path
 import sqlite3
-from typing import Dict, Iterator, Optional, Sequence, Union
+from typing import Dict, Iterator, Sequence, Union
 
 
 class WorkingStateException(Exception):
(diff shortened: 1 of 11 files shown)


source/fab/database.py
```diff
@@ -22,8 +22,11 @@ def __init__(self, filename: Path, adler32: int):
         self.adler32 = adler32
 
     def __eq__(self, other):
+        if not isinstance(other, FileInfo):
+            message = f"Cannot compare FileInfo with {other.__class__}"
+            raise ValueError(message)
         return (str(other.filename) == str(self.filename)) \
```
stevewardle, line 28: I realise this isn't strictly related to this change, but while I was checking this part it occurred to me... if `other.filename` and `self.filename` are identical, aren't we _literally_ talking about the exact same file on disk?  In which case would we ever expect their `adler32` hashes to be different?

source/fab/language/__init__.py
```diff
@@ -33,8 +33,8 @@ def products(self) -> List[Path]:
         raise NotImplementedError('Abstract methods must be implemented')
 
 
-class Analyser(Task):
-    def __init__(self, reader: TextReader, database: SqliteStateDatabase):
+class Analyser(Task, ABC):
```
stevewardle, line 36: Interesting... I thought because `Task` is itself an `ABC` then the behaviour was somehow implicit?  So is this for clarity or does it actually change the behaviour?  (I'm happy with it, just curious)

unit-tests/application_test.py
```diff
@@ -0,0 +1,11 @@
...
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+# We do not unit test the applications as they are far too complex for it.
+# They would require too many fixtures and mocks to test. At that point our
+# unit test would have become a system test.
+#
+# TODO: This is probably an indication that the application classes contain
+#       too much stuff.
```
stevewardle, line 11: Just to say I agree with this assessment - I probably should have added the new "compilation" step within its own new module.  Something for a future ticket for sure

unit-tests/entry_test.py
```diff
@@ -0,0 +1,8 @@
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+# We do not unit test the application entry points as they are far too complex
+# for it. They would require too many fixtures and mocks to test. At that
+# point our unit test would have become a system test.
```
stevewardle, line 8: Why do we have the test files if we aren't going to test anything?  I suppose it just signals to anyone reading the source that we have intentionally not included anything for the corresponding module?

source/fab/database.py
MatthewHambley: That is true if they are referring to the same file at the same instant in time. However if they refer to the same file at different points in time they could have different contents.

source/fab/language/__init__.py
MatthewHambley: I'm not sure what the word of God is. Mypy and flake8 are both happy for abstractness to be inherited but PyCharm complains that `Analyser` does not implement all abstract methods, which of course it doesn't.

I think for clarity it is good to be explicit that this class is also abstract although I am open to the argument that the multiple inheritance approach taken here is inferior to an alternative.

unit-tests/entry_test.py
MatthewHambley: Exactly this. It short circuits people demanding to know why there are no tests for this module. It's like "This page intentionally left blank."

source/fab/database.py
stevewardle: Ah okay I think I see what you're saying - it's dependent upon when exactly the file instances were created (and their hashes computed)

source/fab/language/__init__.py
stevewardle: I think if the behaviour isn't affected by it then it doesn't hurt at all to be explicit.  And especially with PyCharm being quite a popular editor we are smoothing things for any other developers who come to work on Fab and use it.  I was really just curious if there was a practical reason for it that I wasn't aware of

MatthewHambley:
Due to a hamhanded rebase this branch is now untrustworthy. See #51 for the new one.

PR #50, opened 2020-04-02
Queue implimentation

### LINKED ISSUE #30: Implement queue manager
- Initially this will run serially

- Needs some sort of termination condition

- Consider usage of `async_io`

GiorgiaLine:
Aims to impliment a multithreaded queue manager to handle the queueing and running of tasks as their prerequisites are met.

Closes #30

### DIFF
--- a/source/fab/application.py
+++ b/source/fab/application.py
@@ -58,9 +58,8 @@ class Fab(object):
             self._command_flags_map[FortranLinker] = (
                 ld_flags.split()
             )
-        if n_procs is None:
-            n_procs = 2
-        self._queue = QueueManager(n_procs)
+        n_workers: int = n_procs - 1
+        self._queue = QueueManager(n_workers)
 
     def run(self, source: Path):
 
--- a/source/fab/application.py
+++ b/source/fab/application.py
@@ -37,7 +37,8 @@ class Fab(object):
                  exec_name: str,
                  fpp_flags: str,
                  fc_flags: str,
-                 ld_flags: str):
+                 ld_flags: str,
+                 n_procs: int):
 
         self._state = SqliteStateDatabase(workspace)
         self._workspace = workspace
@@ -57,8 +58,9 @@ class Fab(object):
             self._command_flags_map[FortranLinker] = (
                 ld_flags.split()
             )
-        # TODO: Need to pass number of workers from the command line
-        self._queue = QueueManager(4)
+        if n_procs is None:
+            n_procs = 2
+        self._queue = QueueManager(n_procs)
 
     def run(self, source: Path):
 
--- a/source/fab/application.py
+++ b/source/fab/application.py
@@ -75,6 +75,8 @@ class Fab(object):
         descender = TreeDescent(source)
         descender.descend(visitor)
 
+        self._queue.check_queue_done()
+
         file_db = FileInfoDatabase(self._state)
         for file in file_db.get_all_filenames():
             info = file_db.get_file_info(file)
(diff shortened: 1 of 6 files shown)


source/fab/application.py
```diff
@@ -56,16 +57,21 @@ def __init__(self,
             self._command_flags_map[FortranLinker] = (
                 ld_flags.split()
             )
+        # TODO: Need to pass number of workers from the command line
```
MatthewHambley, line 60: Ideally the default behaviour would be "numer of cores on the machine" or some fraction of that. Being able to override that with the command line also makes sense.

source/fab/application.py
```diff
@@ -87,6 +93,8 @@ def run(self, source: Path):
                 print('    found in: ' + str(filename))
                 print('    depends on: ' + str(fortran_db.depends_on(unit)))
 
+        self._queue.check_queue_done()
```
MatthewHambley, line 96: Even though we don't expect to keep the debug output waiting for the queue to empty should probably happen before it.

source/fab/application.py
```diff
@@ -149,34 +157,25 @@ def run(self, source: Path):
...
+            # Add the object files to the linker
+            # TODO: For right now products is a list, though the
+            #       compiler only ever produces a single entry;
+            #       maybe add_object should accept Union[Path, List[Path]]?
+            link_command.add_object(compiler.products[0])
 
         # Hopefully by this point the list is exhausted and
         # everything has been compiled, and the linker is primed
         linker = CommandTask(link_command)
         # TODO: Like the others, this would go on the queue
```
MatthewHambley, line 173: Seems like it does go on the queue.

unit-tests/queue_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+    q_manager = QueueManager(2)
+    q_manager.run()
+
+    for i in range(1, 4):
+        filename = tmp_path / f"file_{i}"
+        task = DummyTask(filename)
+        q_manager.add_to_queue(task)
+
+    q_manager.check_queue_done()
```
MatthewHambley, line 39: I think the check for files should happen at this point. After all the contract of "queue done" is that all owkr is complete.

unit-tests/queue_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for i in range(1, 4):
+        filename = tmp_path / f"file_{i}"
+        assert filename.exists()
+
+
+def test_startstop():
+    q_manager = QueueManager(1)
+    q_manager.run()
+    q_manager.shutdown()
```
MatthewHambley, line 50: How does this check that the shutdown has actually happened? Is there a way of testing that the queue manager is no longer running?

source/fab/application.py
GiorgiaLine: True, done.

unit-tests/queue_test.py
GiorgiaLine: Good point, done.

source/fab/application.py
GiorgiaLine: Missed this in cleanup! Done.

source/fab/application.py
GiorgiaLine: The number of cores on the machine is now the upper limit for what users can select, but I've not made it the default to avoid hogging cores unnessasarily. For safety on weaker systems, the default is set to 2 (the minimum needed to run the main process and one worker), which is what fcm make does currently. I'veaddedan option to override this at the command line.

unit-tests/queue_test.py
GiorgiaLine: This test has changed so should now check the queue manager is no longer running prooperly.

GiorgiaLine:
I believe I've responded to everything now if you'd like to take another look.

PR #53, opened 2020-04-08
Refactor database

MatthewHambley:
This change is quite extensive but is simple in concept. The idea is for the database to represent its contents in a more objecty fashion. Rather than passing back tuples of values or disociated lists it provides nicely packaged objects.


source/fab/language/fortran.py
```diff
@@ -31,10 +30,61 @@
 from fab.reader import TextReader, TextReaderDecorator
 
 
+class FortranUnitUnresolvedID(object):
+    def __init__(self, name: str):
+        self.name = name
+
+    def __eq__(self, other):
+        if not isinstance(other, FortranUnitUnresolvedID):
+            message = "Cannot compare FortranPartialUnitID with " \
```
stevewardle, line 39: Should this message say `FortranUnitUnresolvedID`?  (feels like an earlier iteration of the name)

source/fab/language/fortran.py
```diff
@@ -49,156 +99,165 @@ def __init__(self, database: StateDatabase):
...
-                   on fortran_unit(filename)'''
+                   on fortran_unit (unit, found_in)'''
         ]
         self.execute(create_unit_table, {})
 
         # Although the current unit will already have been entered into the
         # database it is not necessarily unique. We may have multiple source
-        # files which define the same unit. Thus it can not be used as a
-        # foreign key.
+        # files which define similarly named units. Thus it can not be used as
```
stevewardle, line 112: Is it _similarly_ or is it the same?  I know this is just a comment but I'm not sure we want the ambiguity (or maybe I'm wrong and "similar" is right - my impression was that we could have a `something_mod` in 2 or more different source files - is it subtly different to that?)

source/fab/language/fortran.py
```diff
@@ -49,156 +99,165 @@ def __init__(self, database: StateDatabase):
...
-
-        It is possible that the same program unit is multiply defined, hence
-        why a list is returned. It would be an error to try linking these into
-        a single executable but that is not a concern for the model of the
-        source tree.
+    def get_program_unit(self, name: str) -> List[FortranInfo]:
+        """
+        Gets the details of program units given their name.
+
+        It is possible that similarly named program units appear in multiple
```
stevewardle, line 204: Reading this I'm now thinking that this is just be being grammatically ignorant - to be similar like the mathematical term "similar" i.e. they are the same, right?

source/fab/language/fortran.py
MatthewHambley: Of course you are right. What you are seeing there is a left over from a previous name for the class.

source/fab/language/fortran.py
MatthewHambley: I was using "similar" in its mathematic sense. i.e. identical. However your are right to point out that is not common usage. I will change it to something less ambiguous.

source/fab/language/fortran.py
MatthewHambley: Just so, however if you can make that confusion others can too. In code which is not inherently heavily mathematical it's a distinction we don't need to make.

MatthewHambley:
That should be all the requested changes made.

PR #54, opened 2020-04-14
Modify FileTextReader so that it opens files on-demand

stevewardle:
As part of trying to make the tree descent multiprocess-safe it has emerged that it would be better if the `FileTextReader` waited to open files until they are actually needed for the first time.  A functionally null change as far as this PR is concerned, but will eventually mean that any worker process calling a `Task` that requires an instance of this class will open its own copy of the file.


stevewardle:
Taking this back for a moment to work on it slightly more

PR #55, opened 2020-04-17
Make the database connection "on demand"

stevewardle:
To try and enable us to utilise the queue for the source tree descent this proposes a way to make the database connection object only open once an Analyser Task needs to make use of it.  This plays into the way `multiprocessing` works as you cannot pass "live" connections to parallel workers.

However, doing things this way comes at a slight cost - there is some overhead in initialising a connection object (rather than re-using one which was previously opened).  The hope is that being able to utilise multiple workers in the source tree descent offers enough of a benefit on its own that any slight extra overhead in the workers having to create connections becomes worth it

### DIFF
--- a/source/fab/database.py
+++ b/source/fab/database.py
@@ -112,9 +112,6 @@ class SqliteStateDatabase(StateDatabase):
 
     def get_connection(self) -> sqlite3.Connection:
         if self._connection is None:
-            if not self._working_directory.exists():
-                self._working_directory.mkdir(parents=True)
-
             self._connection \
                 = sqlite3.connect(str(self._working_directory / 'state.db'))
             self._connection.row_factory = sqlite3.Row
--- a/source/fab/database.py
+++ b/source/fab/database.py
@@ -104,19 +104,25 @@ class SqliteStateDatabase(StateDatabase):
     '''
     def __init__(self, working_directory: Path):
         self._working_directory: Path = working_directory
-
-        if not self._working_directory.exists():
-            self._working_directory.mkdir(parents=True)
-
-        self._connection: sqlite3.Connection \
-            = sqlite3.connect(str(working_directory / 'state.db'))
-        self._connection.row_factory = sqlite3.Row
+        self._connection: Optional[sqlite3.Connection] = None
 
     def __del__(self):
-        self._connection.close()
+        if self._connection is not None:
+            self._connection.close()
+
+    def get_connection(self) -> sqlite3.Connection:
+        if self._connection is None:
+            if not self._working_directory.exists():
+                self._working_directory.mkdir(parents=True)
+
+            self._connection \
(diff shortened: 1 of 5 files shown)


MatthewHambley:
Can we do some before and after tests on a sizeable source tree to quantify any perforamnce cost?

stevewardle:
> Can we do some before and after tests on a sizeable source tree to quantify any perforamnce cost?

I'll see what I can do - perhaps I can script the generation of a big set of source files that depend on one another in a semi-random (but non-circular) way

stevewardle:
Added the following page also describing some benchmarking work:
https://github.com/metomi/fab/wiki/Database-Connection-Test

source/fab/database.py
```diff
@@ -104,27 +104,30 @@ class SqliteStateDatabase(StateDatabase):
...
-            = sqlite3.connect(str(working_directory / 'state.db'))
-        self._connection.row_factory = sqlite3.Row
+        self._connection: Optional[sqlite3.Connection] = None
 
     def __del__(self):
-        self._connection.close()
+        if self._connection is not None:
+            self._connection.close()
+
+    def get_connection(self) -> sqlite3.Connection:
```
MatthewHambley, line 113: I don't think we should be spilling sqlite objects such as the connection out of this class. I know that Python doesn't actually support data hiding but we can at least pretend it does by preceding the method name with an underline.

unit-tests/database_test.py
```diff
@@ -79,7 +79,15 @@ def test_iteration(self):
 
 class TestSQLiteStateDatabase(object):
     def test_constructor(self, tmp_path: Path):
-        _ = SqliteStateDatabase(tmp_path)
+        database = SqliteStateDatabase(tmp_path)
+
+        assert database._connection is None
+        assert database._working_directory == tmp_path
```
MatthewHambley, line 85: Very uneasy with dipping into the classes "private" variables for testing. Test units should be treated as a black box.

source/fab/database.py
stevewardle: Agreed, I'm actually not sure why I didn't do that to begin with

unit-tests/database_test.py
stevewardle: Does that suggest that we ought to drop testing of the constructor completely following this change then?  As post change I don't think there are any testable effects of the constructor?

stevewardle:
Ready for another look I think - hopefully I've understood what you were getting at with the testing?

unit-tests/database_test.py
MatthewHambley: Testing can be hard. I have checked the coverage and your new tests at last visit every part of the SQLite code. As such I am happy for them to go ahead. We may want to revisit this issue later but that is for a new ticket.

PR #56, opened 2020-04-17
Enable queueing in the tree descent

### LINKED ISSUE #27: Make the tree descent process utilise multiple threads/processes if available

stevewardle:
Followup (and branch currently contains) #54 and #55 to actually make the source tree descent utilise the queue.  The missing piece of the puzzle being that the section which previously ensured that files passing through the descender were hashed can no longer live there (since that loop is now running asynchronously - it would try to hash files that don't yet exist).  The solution is to make the hashing part into another Task which gets added to the queue and depends on the main file whose task is being added.

Doing this also complicates the unit testing of the tree descent slightly - our existing tests were using locally defined classes (i.e. within the existing test routines) and were making use of a variable local to the text but external to those classes.  None of this works now because `multiprocessing` can't serialise locally defined classes, meaning they need to be global again, and making the external array global also fails due to being private to the worker executing the task.  Luckily the module provides a shared memory object that we can use to bypass this

Closes #52 and closes #27

### DIFF
--- a/source/fab/language/__init__.py
+++ b/source/fab/language/__init__.py
@@ -9,8 +9,8 @@ from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import List
 
-from fab.database import StateDatabase
-from fab.reader import TextReader
+from fab.database import StateDatabase, FileInfoDatabase
+from fab.reader import TextReader, TextReaderAdler32
 
 
 class TaskException(Exception):
@@ -35,8 +35,8 @@ class Task(ABC):
 
 class Analyser(Task, ABC):
     def __init__(self, reader: TextReader, database: StateDatabase):
-        self._database = database
         self._reader = reader
+        self._database = database
 
     @property
     def database(self):
@@ -46,6 +46,30 @@ class Analyser(Task, ABC):
     def prerequisites(self) -> List[Path]:
         if isinstance(self._reader.filename, Path):
             return [self._reader.filename]
+        else:
+            return []
+
+    @property
+    def products(self) -> List[Path]:
+        return []
+
+
+class HashCalculator(Task):
+    def __init__(self, hasher: TextReaderAdler32, database: StateDatabase):
+        self._hasher = hasher
+        self._database = database
+
+    def run(self):
+        for _ in self._hasher.line_by_line():
+            pass  # Make sure we've read the whole file
+        file_info = FileInfoDatabase(self._database)
+        file_info.add_file_info(Path(self._hasher.filename),
+                                self._hasher.hash)
+
+    @property
+    def prerequisites(self) -> List[Path]:
+        if isinstance(self._hasher.filename, Path):
+            return [self._hasher.filename]
         else:
             return []
 
--- a/source/fab/source_tree.py
+++ b/source/fab/source_tree.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from typing import Mapping, List, Union, Type
 
 from fab.database import SqliteStateDatabase
-from fab.language import \
+from fab.tasks import \
     Task, \
     Analyser, \
     CommandTask, \
--- a/source/fab/source_tree.py
+++ b/source/fab/source_tree.py
@@ -8,7 +8,7 @@ Descend a directory tree or trees processing source files found along the way.
 '''
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Mapping, List, Union, Type
+from typing import Mapping, List, Union, Type, Callable
 
 from fab.database import SqliteStateDatabase
 from fab.tasks import \
@@ -20,7 +20,6 @@ from fab.tasks.common import \
     CommandTask, \
     HashCalculator
 from fab.reader import TextReader, FileTextReader
-from fab.queue import QueueManager
 
 
 class TreeVisitor(ABC):
@@ -35,12 +34,12 @@ class ExtensionVisitor(TreeVisitor):
                  command_flags_map: Mapping[Type[Command], List[str]],
                  state: SqliteStateDatabase,
                  workspace: Path,
-                 queue: QueueManager):
+                 task_handler: Callable):
         self._extension_map = extension_map
         self._command_flags_map = command_flags_map
         self._state = state
         self._workspace = workspace
-        self._queue = queue
+        self._task_handler = task_handler
 
     def visit(self, candidate: Path) -> List[Path]:
         new_candidates: List[Path] = []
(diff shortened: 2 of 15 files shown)


stevewardle:
Closes #27 and closes #52

source/fab/language/__init__.py
```diff
@@ -9,8 +9,8 @@
 from pathlib import Path
 from typing import List
 
-from fab.database import StateDatabase
-from fab.reader import TextReader
+from fab.database import StateDatabase, FileInfoDatabase
+from fab.reader import TextReader, TextReaderAdler32
```
MatthewHambley, line 13: I'm not sure I like having `FileInfoDatabase` in `fab.database`. What do you think about a new module which contains both it and `TextReaderAdler32`? That way the two components which work together live together and neither sit in their parents modules.

source/fab/language/__init__.py
```diff
@@ -54,6 +54,30 @@ def products(self) -> List[Path]:
...
+        if isinstance(self._hasher.filename, Path):
+            return [self._hasher.filename]
+        else:
+            return []
+
+    @property
+    def products(self) -> List[Path]:
+        return []
+
+
```
MatthewHambley, line 80: I wonder if we're putting too much int `__init__.py`. It's a magic file, we probably shouldn't be over using it.

source/fab/source_tree.py
```diff
@@ -57,16 +58,12 @@ def visit(self, candidate: Path) -> List[Path]:
                 message = \
                     f'Unhandled class "{task_class}" in extension map.'
                 raise TypeError(message)
-            # TODO: Make SQLite connection multiprocess safe
-            # self._queue.add_to_queue(task)
-            task.run()
+
+            self._queue.add_to_queue(task)
+            self._queue.add_to_queue(HashCalculator(hasher, self._state))
```
MatthewHambley, line 63: One advantage of the hasher as a separate task is that you don't have to worry about any previous actions performed by the task stack altering the file you're hashing.

unit-tests/source_tree_test.py
```diff
@@ -12,6 +12,27 @@
 from fab.source_tree import ExtensionVisitor, TreeDescent, TreeVisitor
 from fab.queue import QueueManager
 
+from multiprocessing import Manager
+
+
+tracker = Manager().list()
```
MatthewHambley, line 18: Could this be a static class variable? i.e. defined at the top level of the class:

```python
class Whatever:
    static_var = 'foo'
```

unit-tests/source_tree_test.py
```diff
@@ -20,81 +41,73 @@ def test_analyser(self, tmp_path: Path):
...
         }
         fmap: Dict[Type[Command], List[str]] = {}
         queue = QueueManager(1)
+        queue.run()
         test_unit = ExtensionVisitor(emap, fmap, db, tmp_path, queue)
-        tracker.clear()
+
+        tracker[:] = []
+
         test_unit.visit(tmp_path / 'test.bar')
```
MatthewHambley, line 82: Is there a way we can implement the visitor so it can be tested on its own, without the queue stuff? A way in which all the queueing happens in the descent and not the visitor?

source/fab/language/__init__.py
stevewardle: I see where you're coming from - as a separate ticket?

source/fab/language/__init__.py
stevewardle: Yeah you may be write, I suppose we could have a `fab.language.generic`?

source/fab/language/__init__.py
```diff
@@ -54,6 +54,30 @@ def products(self) -> List[Path]:
...
+    def __init__(self, hasher: TextReaderAdler32, database: StateDatabase):
+        self._hasher = hasher
+        self._database = database
+
+    def run(self):
+        for _ in self._hasher.line_by_line():
+            pass  # Make sure we've read the whole file
+        file_info = FileInfoDatabase(self._database)
+        file_info.add_file_info(Path(self._hasher.filename),
+                                self._hasher.hash)
```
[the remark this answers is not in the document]
stevewardle: I may be mistaken in thinking this but I thought by doing it in this way it would preserve the original intended behaviour (i.e. if the hashing reader has already read the file it won't re-read it just to create the hash)... though now that I say that perhaps by putting these on the queue that no longer happens anyway?

unit-tests/source_tree_test.py
stevewardle: I thought I tried something like this and it didn't like it but I will have another go

unit-tests/source_tree_test.py
stevewardle: We could go back to a model where the visitor `yield`s stuff for the descender to deal with (only this time it would be `yield`ing the task that needs to be stuck on the queue)... though we've also got the hash calculator to add which might complicate things

source/fab/language/__init__.py
stevewardle: I've left this alone for now - in part because we've actually now _lost_ `TextReaderAdler32` owing to one of the comments further down.  We could move `FileInfoDatabase` into its own module but I'm a little unsure

source/fab/language/__init__.py
stevewardle: Following some discussion, `language` has been renamed to `tasks` and we now have `fab.tasks.common` which contains `HashCalculator` and `CommandTask` - the rest has stayed in `__init__.py`

source/fab/language/__init__.py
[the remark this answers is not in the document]
stevewardle: This is indeed the case - so following some offline discussion I've done as suggested - the `TextReader` decorator that did the hashing before is gone and its logic has been migrated directly into the `HashCalculator` task

unit-tests/source_tree_test.py
stevewardle: In the end I've gone for changing the visitor so it accepts a `Callable` in its constructor that it will call with any `Task` instance it wishes to run - this way the queuing mechanism is pulled up into the application layer (and we can bypass it in the unit tests)

unit-tests/source_tree_test.py
stevewardle: This was rendered moot by the change lower down (since the unit tests no longer care about the queue)

stevewardle:
Okay I think I've addressed all the points and this is ready for a repeat viewing

source/fab/source_tree.py
```diff
@@ -33,40 +34,35 @@ def __init__(self,
...
             else:
                 message = \
                     f'Unhandled class "{task_class}" in extension map.'
                 raise TypeError(message)
-            # TODO: Make SQLite connection multiprocess safe
-            # self._queue.add_to_queue(task)
-            task.run()
+
+            self._task_handler(task)
+            self._task_handler(HashCalculator(reader, self._state))
```
MatthewHambley, line 62: I don't think we should be hard wiring tasks into the descender. It should probably be for the layer above to decide whether files need to be hashed or not.

MatthewHambley: After discussion with the developer we have agreed to go ahead with the change as it stands and reconsider the whole issue of `ExtensionVisitor` in ticket #58.

MatthewHambley: Making this change is postponed to #58

PR #72, opened 2020-06-02
Localise entry points

### LINKED ISSUE #62: Create top level framework for Extraction Tool
The first step to create our extraction tool - `fab-grab` is to add the necessary top level structure (i.e. the entries in `application.py` and supporting things in the setup, command line interface etc.) so that it can be called

MatthewHambley:
This change alters how we deal with entry points. It does away with the entry module and instead each application module provides its own entry point.

This should make things more self contained. Adding a new application doesn't require editing an existing file other than the list of entry points in setup.

There's also a little light code tidying in here as well.

This is a follow up to #70 and closes #62 .

This replaces and closes #71.


setup.py
```diff
@@ -44,15 +44,15 @@
...
     package_dir={'': 'source'},
     packages=setuptools.find_packages(where='source'),
-    entry_points={'console_scripts': ['fab=fab.entry:fab_entry',
-                                      'fab-grab=fab.entry:grab_entry',
-                                      'fab-dump=fab.entry:dump_entry'],
-                  'gui_scripts': ['fab-explorer=fab.entry:fab_explorer']},
+    entry_points={'console_scripts': ['fab=fab.builder:entry',
+                                      'fab-grab=fab.grabber:entry',
+                                      'fab-dump=fab.dumper:entry'],
+                  'gui_scripts': ['fab-explorer=fab.explorer:entry']},
```
stevewardle, line 50: What a strange coincidence that the equals happen to line up here!

source/fab/builder.py
```diff
@@ -27,6 +26,72 @@
...
 
+def entry() -> None:
+    """
+    Entry point for the Fab build tool.
+    """
+    import argparse
+    import multiprocessing
+    import sys
+    import fab
+
```
stevewardle, line 37: This looks strange to me - I'm surprised `flake8` is okay with it?  Certainly `pylint` will throw an absolute wobbly if you have `import`s that aren't at the top level.  What is the advantage of them being here?

source/fab/dumper.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def entry() -> None:
+    """
+    Entry point for the Fab state database dump tool.
+    """
+    import argparse
+    import fab
+
+    logger = logging.getLogger('fab-dumper')
```
stevewardle, line 24: There's a bit of mixing of naming, the entry point is referred to as `fab-dumper` in `setup.py` and in the source filename of this new file, but the actual command is called `fab-dump`.  I wonder if perhaps from the user's point of view it makes more sense for the logging identifier to match the name of the command they typed.  (On the other hand, it's not as if it's ambiguous; there is clearly only one command it could be talking about)

source/fab/grabber.py
```diff
@@ -0,0 +1,65 @@
...
+    if arguments.verbose:
+        logger.setLevel(logging.INFO)
+    else:
+        logger.setLevel(logging.WARNING)
+
+        application = Grab(arguments.destination)
+
+        repositories = [repository_from_url(url)
+                        for url in arguments.repositories]
+        application.run(repositories)
```
stevewardle, line 56: This block looks like it is accidentally in the `else` clause for the `logger`

source/fab/grabber.py
```diff
@@ -0,0 +1,65 @@
...
+
+def entry() -> None:
+    """
+    Extract and merge up a source tree from several repositories.
+    """
+    import argparse
+    import sys
+    import fab
+
+    logger = logging.getLogger('fab-grab')
```
stevewardle, line 24: Related to my earlier comment about `fab-dumper` - the logging here uses the name of the command-line tool rather than being `fab-grabber` (personally I think I prefer this way; so it corresponds to the name of the typed command)

source/fab/repository.py
```diff
@@ -0,0 +1,75 @@
...
+        if report.returncode != 0:
+            message = f"Unable to extract Subversion repository: {self.url}"
+            raise FabException(message)
+
+
+def repository_from_url(url: str) -> Repository:
+    """
+    Creates an appropriate Repository object from a given URL.
+
+    TODO: This will need to be considerably more illaborate in the future once
```
stevewardle, line 66: elaborate (sorry for pedantry!)

system-tests/test.py
```diff
@@ -19,118 +20,124 @@
...
+        self.standard_out: Optional[str] = None
+        self.standard_error: Optional[str] = None
+
+    @property
+    def test_parameters(self) -> TestParameters:
+        return self._parameters
+
+    @abstractmethod
+    def description(self) -> str:
+        raise NotImplementedError("Abstract methods must be implemented.")
```
stevewardle, line 73: I think the wording of the `NotImplementedError` here is inconsistent with one further back in the changelog (and potentially with some of the ones elsewhere in the codebase - minor thing but can you check and make sure they keep to the same text?)

source/fab/builder.py
MatthewHambley: Importing is bound to scope like everything else. That means these libraries are only visible in the subroutine they are imported to. I chose to do this as they aren't needed anywhere else. It seemed neater.

source/fab/dumper.py
MatthewHambley: You've identified a problem but it's the opposite to the one you think it is.

loggers are named in a dot separated hierarchy and messages are passed up the tree until something processes them. When the logger is used it is fetched using `logging.getLogger(__name__)`, i.e. using the package name.

So in order for any logging to ever happen the logger which is set up in the entry point must be on that hierarchy. In other words, it must be named `fab`.

source/fab/grabber.py
MatthewHambley: Yes it does.

system-tests/test.py
MatthewHambley: This was right, the repository was wrong.

setup.py
```diff
@@ -44,7 +44,8 @@
         ],
     package_dir={'': 'source'},
     packages=setuptools.find_packages(where='source'),
-    entry_points={'console_scripts': ['fab=fab.builder:entry',
+    entry_points={'console_scripts': ['fab=fab.builder:entry',  # Alias
+                                      'fab-build=fab.buioder:entry',
```
stevewardle, line 48: Typo on `builder` here

source/fab/dumper.py
stevewardle: Ah okay!  That makes sense

PR #76, opened 2020-06-25
Simple extract from Git

### LINKED ISSUE #64: Implement extraction from Git repository

MatthewHambley:
This change implements a simple Git extraction capability. It can't cope with branches or any of that good stuff but it's a start.

Due to our system apparently not having the Git daemon installed the testing is limited to file based repositories.

I don't know what's happend with all the commits "Initial import" which I didn't make. It may be GitHub getting upset because there's a Git repository in the repository.

Closes #64.


MatthewHambley:
Looks like this branch is broken.

PR #77, opened 2020-06-25
Simple git extraction

### LINKED ISSUE #64: Implement extraction from Git repository

MatthewHambley:
This branch implements basic extraction from a Git repository. It doesn't handle branches or any of that good stuff but it's a start.

Closes #64 .

This is the second attempt at this branch. The previous one went mental. I think it may have been to do with storing a Git repository in a Git rpeository. As such we now store our test repositories in tarballs.


source/fab/repository.py
```diff
@@ -12,8 +12,9 @@
 """
 from abc import ABC, abstractmethod
 from pathlib import Path
-from subprocess import run
-from urllib.parse import urlparse
+from subprocess import PIPE, run, Popen
+import tarfile
+from urllib.parse import urlparse, urlunparse
```
stevewardle, line 17: I think this could be another case for our style guidelines or conventions - I sometimes feel like the `from x import y` statements shouldn't be mixed with `import z` as it looks a bit strange/messy.  On the other hand - could you be explicit and do `from tarfile import ?` anyway?

source/fab/repository.py
```diff
@@ -38,6 +39,38 @@ def extract(self, target: Path):
...
+        process = Popen(command, stdout=PIPE)
+
+        extractor = tarfile.open(fileobj=process.stdout, mode='r|')
+        extractor.extractall(target)
+
+        process.wait(self._TIMEOUT)
+        if process.returncode != 0:
+            message = "Fault exporting tree from Git repository."
+            if process.stderr:
+                message += '\n' + str(process.stderr)
```
stevewardle, line 70: I'm not sure that `process.stderr` will contain anything unless you pass `stderr=PIPE` to `Popen` (maybe I'm wrong though?)  Also does the message work correctly here or would it need to be something like `process.stderr.decode()`?)

system-tests/test.py
```diff
@@ -227,23 +228,26 @@ class RunGrab(EnterPython):
...
+            #       which is way too much to consider at the moment.
             # repo_url = f'http://localhost/repo'
-            raise Exception("Unable to test Subversion over HTTP protocol.")
+            message = "Unable to test Fetch over HTTP protocol."
+            raise NotImplementedError(message)
+
+        repo_url = f'{self._scheme}://'
+        if protocol == 'file':
+            repo_url += f'//{self._repo_path}'
+        elif protocol in ['git', 'http', 'svn']:
```
stevewardle, line 247: Perhaps `http` shouldn't be included here if it is excepted just above?

unit-tests/repository_test.py
```diff
@@ -90,18 +93,126 @@ def test_extract_from_http(self, repo: Tuple[Path, Path], tmp_path: Path):
...
+        repo_path = tmp_path_factory.mktemp('repo', numbered=True)
+        command = ['git', 'init', str(repo_path)]
+        assert run(command).returncode == 0
+        #
+        # We have to configure this information or the forthcoming commands
+        # will fail.
+        #
+        command = ['git', 'config', 'user.name', 'Testing Tester Tests']
+        assert run(command, cwd=str(repo_path)).returncode == 0
+        command = ['git', 'config', 'user.email', 'tester@example.com']
```
stevewardle, line 120: I mentioned this too you offline - but I think the testing may need to restore the user's original settings or something because I think you have _become_ `Testing Tester Tests` since this part first ran under your account

source/fab/repository.py
MatthewHambley: It would probably be a bad idea since what I am importing from `tarfile` is the function `open`. Obviously this would clash with the intrinsic open function.

I vascilate on how best to handle imports. For classes I like to import them seperately so I don't have to give full paths when referring to them. They tend to have reasonably unique names so collisions are not such a problem.

For global functions I prefer to import them through the module since they often have generic names like `open` or `run`.

Unfortunately it is hard to work out a good way to do both, see the subprocess line for an example. Ideally I would import just the Popen class on its own and the PIPE and run functions in their namespace.

stevewardle: Ah yes, I see what you mean... I guess you could do `from tarfile import open as taropen` but that introduces your own naming decision into the mix, further obfuscating things.  For the `subprocess` case you could always do it on more than one line?  (have both `import subprocess` and `from subprocess import Popen`) - but again maybe that is just as bad (or worse!)

MatthewHambley: I don't know. This is why I vascilate.

source/fab/repository.py
MatthewHambley: The documentation is unclear on this subject so I tried it out and it looks like you are right.

system-tests/test.py
MatthewHambley: While the code is correct, HTTP would require a hostname, I don't care strongly enough to argue. I've removed it.

unit-tests/repository_test.py
MatthewHambley: You are right and in a sphincter tightening manouver I updated those commits with the correct information. I'm not sure that a version control system should allow that sort of monkey business so easily but it was useful in this case.

MatthewHambley:
Unfortunately fixing the reposity seems to have broken the pull request so I'll have to raise a new one.

PR #78, opened 2020-06-30
Simple extraction from a Git repository

### LINKED ISSUE #64: Implement extraction from Git repository

MatthewHambley:
There are many limitations, not least of which that it cannot extract from a branch. It should get us started however.

Closes #64


MatthewHambley:
Smart move. It's a bit hairy having nested repositories like this. Can't see any other way to test it though.

stevewardle:
Happy to say that this didn't do anything strange when checked out locally, so the changes made to avoid affecting the user's identity are all working as intended (y)

PR #79, opened 2020-07-08
Rework extension visitor to use pattern matching rather than extension

stevewardle:
As part of preparing to support the slightly more complicated workflow involved in the pre-compilation steps for C source and header files, this updates the extension visitor so that instead of doing a simple check on the extension of a file, it defines and matches a regex.  In this way it will be possible to have multiple steps using different Task classes on files with the same extension (but different paths) in a way that means we don't tie ourselves in knots inventing many bespoke intermediate extensions

### DIFF
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -6,8 +6,7 @@
 from collections import defaultdict
 import logging
 from pathlib import Path
-from typing import Dict, List, Type, Union, Tuple
-from re import match
+from typing import Dict, List, Type
 
 from fab import FabException
 from fab.database import SqliteStateDatabase, FileInfoDatabase
@@ -23,7 +22,10 @@ from fab.tasks.fortran import \
     FortranUnitID, \
     FortranCompiler, \
     FortranLinker
-from fab.source_tree import TreeDescent, SourceVisitor
+from fab.source_tree import \
+    TreeDescent, \
+    SourceVisitor, \
+    PathMap
 from fab.queue import QueueManager
 
 
@@ -94,13 +96,14 @@ def entry() -> None:
 
 
 class Fab(object):
-    _precompile_map: List[Tuple[str, Union[Type[Task], Type[Command]]]] = [
+    _precompile_map = PathMap([
         (r'.*\.f90', FortranAnalyser),
         (r'.*\.F90', FortranPreProcessor),
-    ]
-    _compile_map: List[Tuple[str, Type[Command]]] = [
+    ])
+
+    _compile_map = PathMap([
         (r'.*\.f90', FortranCompiler),
-    ]
+    ])
 
     def __init__(self,
                  workspace: Path,
@@ -222,14 +225,9 @@ class Fab(object):
             #       expect to be *produced* by the compile
             #       and pass this to the constructor for
             #       inclusion in the task's "products"
-            compiler_class = None
-            for pattern, classname in self._compile_map:
-                # Note we keep searching through the map
-                # even after a match is found; this means that
-                # later matches will override earlier ones
-                if match(pattern, str(unit.found_in)):
-                    compiler_class = classname
-
+            compiler_class = (
+                self._compile_map.get_task(unit.found_in)
+            )
             if compiler_class is None:
                 continue
 
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -6,7 +6,7 @@
 from collections import defaultdict
 import logging
 from pathlib import Path
-from typing import Dict, List, Type, Union
+from typing import Dict, List, Type, Union, Tuple
 from re import match
 
 from fab import FabException
@@ -94,13 +94,13 @@ def entry() -> None:
 
 
 class Fab(object):
-    _precompile_map: Dict[str, Union[Type[Task], Type[Command]]] = {
-        r'.*\.f90': FortranAnalyser,
-        r'.*\.F90': FortranPreProcessor,
-    }
-    _compile_map: Dict[str, Type[Command]] = {
-        r'.*\.f90': FortranCompiler,
-    }
+    _precompile_map: List[Tuple[str, Union[Type[Task], Type[Command]]]] = [
+        (r'.*\.f90', FortranAnalyser),
+        (r'.*\.F90', FortranPreProcessor),
+    ]
+    _compile_map: List[Tuple[str, Type[Command]]] = [
+        (r'.*\.f90', FortranCompiler),
+    ]
 
     def __init__(self,
                  workspace: Path,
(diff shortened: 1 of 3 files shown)


source/fab/builder.py
```diff
@@ -93,13 +94,13 @@ def entry() -> None:
...
-    _compiler_map: Dict[str, Type[Command]] = {
-        '.f90': FortranCompiler,
-    }
+    _precompile_map: List[Tuple[str, Union[Type[Task], Type[Command]]]] = [
+        (r'.*\.f90', FortranAnalyser),
+        (r'.*\.F90', FortranPreProcessor),
+    ]
+    _compile_map: List[Tuple[str, Type[Command]]] = [
+        (r'.*\.f90', FortranCompiler),
+    ]
```
MatthewHambley, line 103: Tuples are often a convenient but ugly solution to problems. Is there a cleaner way of doing this?

Would a map of string to class be better? I imagining ordering is important and random look-up isn't.

Would a simple class with named properties make for a nicer solution? In such a case the compilation of the RegEx could be included in that object rather than having to do it later.

source/fab/builder.py
```diff
@@ -221,7 +222,16 @@ def run(self, source: Path):
             #       expect to be *produced* by the compile
             #       and pass this to the constructor for
             #       inclusion in the task's "products"
-            compiler_class = self._compiler_map[unit.found_in.suffix]
+            compiler_class = None
+            for pattern, classname in self._compile_map:
+                # Note we keep searching through the map
+                # even after a match is found; this means that
+                # later matches will override earlier ones
+                if match(pattern, str(unit.found_in)):
```
MatthewHambley, line 230: I note the match is being done from text each time. This will be considerably slower than using a compiler pattern however that can be an optimisation for later.

source/fab/source_tree.py
```diff
@@ -27,42 +28,50 @@ def visit(self, candidate: Path) -> List[Path]:
...
     def __init__(self,
-                 extension_map: Mapping[str, Union[Type[Task], Type[Command]]],
+                 source_map:
+                     List[Tuple[str, Union[Type[Task], Type[Command]]]],
                  command_flags_map: Mapping[Type[Command], List[str]],
                  state: SqliteStateDatabase,
                  workspace: Path,
                  task_handler: Callable):
-        self._extension_map = extension_map
+        self._source_map = source_map
```
MatthewHambley, line 39: I'm not sure I like "source map". Would something like "path map" more readily describe what is happening?

source/fab/source_tree.py
```diff
@@ -27,42 +28,50 @@ def visit(self, candidate: Path) -> List[Path]:
...
-        except KeyError:
-            pass
+
+        task_class = None
+        for pattern, classname in self._source_map:
+            # Note we keep searching through the map
+            # even after a match is found; this means that
+            # later matches will override earlier ones
+            if match(pattern, str(candidate)):
+                task_class = classname
```
MatthewHambley, line 54: I smell duplicate code. Is something wrong here?

source/fab/builder.py
stevewardle: Yes I think this is an obvious area for improvement - you'll see in the upcoming PR that implements "phases" that this becomes slightly uglier still (for the other map just above this).  I expect the solution is as you suggest to turn a "phase" or a "map" into an object with named properties.  But I think this might be easier after the aforementioned PR and this are both committed (what do you think?)

source/fab/builder.py
stevewardle: One thing to keep in mind (and related to the above comment too) is that (eventually, in the wider context of what this is going towards with the "phases") the patterns end up containing substitutions as well.  e.g. the pattern can contain `{source}` which gets substituted for the source path of the current phase (and this happens at the start of each phase).  But perhaps that is the perfect place to eventually stick a call to `compile` the pattern (it may need an intermediate object for that, but maybe that can be part of the "phase" / "map" object discussed above

source/fab/source_tree.py
stevewardle: Yes that makes sense, source is probably best avoided

source/fab/source_tree.py
stevewardle: A candidate for the phase / map object as well I expect (it can have a `.select_class` method or something which does this)

source/fab/builder.py
stevewardle: In the end I had a go at making these into an object, which at least hides a bit of the ugliness

source/fab/source_tree.py
stevewardle: I put this into the new object I created in the end so the duplication is gone

source/fab/source_tree.py
```diff
@@ -27,42 +27,58 @@ def visit(self, candidate: Path) -> List[Path]:
         raise NotImplementedError("Abstract method must be implemented")
 
 
-class ExtensionVisitor(TreeVisitor):
+class PathMap(object):
```
MatthewHambley, line 30: I think ultimately the PathMap class should represent a single mapping and the list should be of those objects. For the moment this is fine.

unit-tests/source_tree_test.py
```diff
@@ -26,15 +30,15 @@ def run(self):
                 tracker.append(self._reader.filename)
 
         db = SqliteStateDatabase(tmp_path)
-        emap: Dict[str, Union[Type[Task], Type[Command]]] = {
-            '.foo': DummyTask
-        }
+        smap = PathMap([
+            (r'.*\.foo', DummyTask),
```
MatthewHambley, line 34: Note that here `.foo` means "any character followed by foo". Is that what you wanted? Did you actually mean `\.foo` which means ".foo"?

unit-tests/source_tree_test.py
```diff
@@ -63,15 +67,15 @@ def as_list(self) -> List[str]:
                 return ['cp', str(self._filename), str(self.output[0])]
 
         db = SqliteStateDatabase(tmp_path)
-        emap: Dict[str, Union[Type[Task], Type[Command]]] = {
-            '.bar': DummyCommand
-        }
+        smap = PathMap([
+            (r'.*\.bar', DummyCommand),
```
MatthewHambley, line 71: Wildcard when literal may have been intended.

source/fab/builder.py
MatthewHambley: As long as you raise an issue for it I am happy to hold off until later.

source/fab/builder.py
MatthewHambley: It may be that the subsitution becomes unecessary down the line but even if we can just compile and use twice we win over two uses.

source/fab/source_tree.py
stevewardle: That does make sense - I will keep it in mind when updating #80 after this (as that will require a little extra work to this area anyway)

unit-tests/source_tree_test.py
stevewardle: The pattern is passed to `re.match` so I think it has to match the entire path (in the new version) - so it's `.*` which should match absolutely anything followed by `\.foo` meaning it must end with extension `.foo` - functionally I think this is identical to the behaviour before since it used to only look at the extension

unit-tests/source_tree_test.py
stevewardle: As above, I think this is the correct pattern to match any path to a file with the `.bar` extension

PR #80, opened 2020-07-15
Phases

### LINKED ISSUE #75: Implement "processing stages"
As I've begun working on handling C source, and with a view to what we will eventually want to support with Psyclone, I'd like to investigate introducing something I'm going to call "processing stages".  The need for this is illustrated by this:

 * For our C source the intention is that we will have to first descend the source tree and inject special `#pragma`s that will allow us to contextualise included headers after pre-processing

 * The problem is that since we will need to do this in both `.c` and `.h` files, and the `.h` files must have been injected with `#pragma`s _before_ the preprocessing step on any `.c` file that includes them - we have another situation similar to the relationship between the preprocessing/analysis and compilation

 * So it feels like what this is leaning towards is doing the injection step, then having a call to flush the queue of scheduled tasks, and _then_ proceed with the preprocessing and analysis steps

My idea for how we might accommodate this is that you would have "stages" which are each represented in the code as an instance of a dictionary mapping extensions onto `Task`s (as we currently have with the `extension_map` and `compiler_map`).  The code can then be structured so that a stage is processed and then the queue is flushed.  So e.g. we might have something like:

```
_analysis_stage_1_map = {
    ".c": CInjectPragmas,
    ".h": CInjectPragmas,
}

_analysis_stage_2_map = {
    ".c": CPreProcess,
    ".c-fab-pp": CAnalyser
}
```
These could in turn be passed to the `ExtensionVisitor` and `TreeDescent` and have it `descend` to process everything it can in a stage, and then flush the queue and move onto the next stage. 

Perhaps in the `working` directory each stage can have its own subdirectory which is where files produced by that stage end up (or something) which allows us to re-use extensions if they are from different stages (as above).

Longer term defining these stages would probably be something which the configuration can setup and/or customise - allowing for a flexible variety of different things to happen as part of a build

stevewardle:
This implements the idea of having multiple "phases" to (for now) the steps leading up to the compilation + linking (so all the preprocessing and analysis type tasks).  It allows several of our `extension_map` dictionaries to be provided - each of these represents a "phase", with a queue synchronisation inbetween each of these (which allows us to mandate types of `Task` which must complete across the entire source tree before the next phase can begin - similar to the sync between the old preprocess/analysis steps and the compilation itself)

In order to manage the output files a new system is implemented in the `working` directory - a `core` directory contains sym links to the "current state" of the build (initially this is setup to link to the original source files).  Each `phase` points at this directory for its input, but places any produced files into its own `phase_?` directory (incrementing each phase).  At the end of each phase the files in the `core` directory are updated to point to any new files in the (just completed) `phase_?` directory.  In this way `Task`s can produce files with identical extensions as output, and it should hopefully lend itself nicely to debugging and implementing incremental builds later.

Closes #75

### DIFF
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -102,10 +102,10 @@ class Fab(object):
 
     _phase_maps = [
         PathMap([
-            (r'.*\.f90', FortranAnalyser),
+            (r'.*\.F90', FortranPreProcessor),
         ]),
         PathMap([
-            (r'.*\.F90', FortranPreProcessor),
+            (r'.*\.f90', FortranAnalyser),
         ]),
     ]
 
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -25,7 +25,8 @@ from fab.tasks.fortran import \
 from fab.source_tree import \
     TreeDescent, \
     SourceVisitor, \
-    PathMap
+    PathMap, \
+    CoreLinker
 from fab.queue import QueueManager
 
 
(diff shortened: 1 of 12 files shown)


stevewardle:
Closing this as it has now been superceded by the "Artifact" implementation

PR #82, opened 2020-07-23
Rework internal processing to use Artifacts and a "state machine" like approach

stevewardle:
Whilst wrestling with the more complicated cases introduced by trying to extend Fab to handle C source it emerged that perhaps the existing approach of instantiating Tasks as part of the descent and having the descent extended by their results was going to struggle with more complex examples.

So after much thought and discussion here we present a completely different approach which moves much of the processing logic inside of the queue system, via a new "Engine" that works like a state machine.  A new Artifact object which represents a file on disk with some attached details becomes the new work unit which passes through the queue and is handled by the Engine.  This also brings the pre and post analysis tasks together so that we no longer have the artifical barrier between analysing the source and beginning compilation.

Unfortunately it's quite a large and destructive change but hopefully worthwhile


stevewardle:
Marking this as ready - I do think it is going to need some further work; the biggest emerging problem with it is going to be handling errors and other situations (as an example - at the moment specifying a non-existent target results in a lockup)

source/fab/artifact.py
```diff
@@ -0,0 +1,113 @@
...
+    @property
+    def hash(self) -> int:
+        # If this is the first access of the property calculate the hash
+        # and cache it for later accesses
+        if self._hash is None:
+            self._hash = 1
+            reader = FileTextReader(self.location)
+            for line in reader.line_by_line():
+                self._hash = adler32(bytes(line, encoding='utf-8'), self._hash)
+        return self._hash
```
MatthewHambley, line 107: On demand hashing - nice.

source/fab/builder.py
```diff
@@ -119,43 +109,67 @@ def __init__(self,
...
-            )
-        self._queue = QueueManager(n_procs - 1)
 
-    def _extend_task_queue(self, task: Task) -> None:
-        self._queue.add_to_queue(task)
-        for prereq in task.prerequisites:
-            self._queue.add_to_queue(HashCalculator(FileTextReader(prereq),
-                                                    self._state))
+        # If not given, use the name of the target as the exec name
+        # TODO: Should this be in the CLI part?
```
MatthewHambley, line 114: I think it probably should. The actual implementation of the business logic can expect to be given exact details of what to do. It is the UIs job to deal with default behavior and stuff like that. Please create a ticket for moving this or do it now if the opportunity presents itself.

source/fab/builder.py
```diff
@@ -119,43 +109,67 @@ def __init__(self,
...
+        fortran_preprocessor = FortranPreProcessor(
+            'cpp', ['-traditional-cpp', '-P'] + fpp_flags.split(), workspace
+        )
+        fortran_analyser = FortranAnalyser(workspace)
+        fortran_compiler = FortranCompiler(
+            'gfortran',
+            ['-c', '-J', str(workspace)] + fc_flags.split(), workspace
+        )
+        # TODO: Linker class is probably generic not Fortran specific
+        # now that we are passing in the command to use?
```
MatthewHambley, line 140: It always was but that was not important when we were only considering Fortran. Not something for this changeset.

source/fab/builder.py
```diff
@@ -119,43 +109,67 @@ def __init__(self,
...
     def run(self, source: Path):
 
         self._queue.run()
 
-        visitor = SourceVisitor(self._precompile_map,
-                                self._command_flags_map,
-                                self._state,
-                                self._workspace,
-                                self._extend_task_queue)
+        visitor = SourceVisitor(self._extend_queue)
```
MatthewHambley, line 167: Yes, reducing number of arguments to the tree descender feels right.

source/fab/builder.py
```diff
@@ -172,91 +186,3 @@ def run(self, source: Path):
...
-            #       compiler only ever produces a single entry;
-            #       maybe add_object should accept Union[Path, List[Path]]?
-            link_command.add_object(compiler.products[0])
-
-        # Hopefully by this point the list is exhausted and
-        # everything has been compiled, and the linker is primed
-        linker = CommandTask(link_command)
-        self._queue.add_to_queue(linker)
-        self._queue.check_queue_done()
-        self._queue.shutdown()
```
MatthewHambley, line 262: Moving all this logic out and giving it a place of its own also feels right.

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+    def filetype(self):
+        return self._filetype
+
+    @property
+    def state(self):
+        return self._state
+
+    def match(self, path: Path) -> bool:
+        matched = False
+        if re.match(self.pattern, str(path)):
```
MatthewHambley, line 52: We should compile this pattern in the constructor for faster matching. Please raise a ticket for this.

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class Engine(object):
+    def __init__(self,
+                 workspace: Path,
+                 target: str,
+                 pathmaps: List[PathMap],
+                 taskmap: Mapping[
+                     Tuple[Type[FileType], Type[State]],
+                     Task]) -> None:
```
MatthewHambley, line 64: The excessivenes of the `taskmap` type information suggests that there may be room for improvement here. Have a think but no need to implement anything in this change.

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+        if (artifact.state is New
+                and artifact.filetype is Unknown):
+            # Use the pathmap list to work out the
+            # filetype and starting state
+            new_artifact = None
+            for pathmap in self._pathmaps:
+                if pathmap.match(artifact.location):
+                    new_artifact = Artifact(artifact.location,
+                                            pathmap.filetype,
+                                            pathmap.state)
```
MatthewHambley, line 94: There is probably a neater way of doing this by implementing `__special__` methods on the artifact. Maybe a way to use the `in` keyword. Something to bare in mind for subsequent changes.

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+            # will have an appropriate task that should
+            # be used to run it (though unlike the old
+            # implementation this is probably returning
+            # the instance of the Task not the class)
+            if ((artifact.filetype, artifact.state)
+                    in self._taskmap):
+                task = self._taskmap[(artifact.filetype,
+                                      artifact.state)]
+
+                new_artifacts.extend(task.run([artifact]))
```
MatthewHambley, line 181: Rather than the big compound `if` tree it may be possible to streamline this using a dictionary. An optimisation for subsequent consideration.

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+                file_info = FileInfoDatabase(self._database)
+                file_info.add_file_info(artifact.location,
+                                        new_artifact.hash)
+                new_artifacts.append(new_artifact)
+
+        elif artifact.state is Analysed:
+
+            # Work out whether this artifact needs to be
+            # included in the build or not - tag it as having
+            # been seen so the system knows it is Analysed
```
MatthewHambley, line 109: Surely if `artifact.sate is Analysed` then it has been analysed?

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+                for idep, dependency in enumerate(artifact.depends_on):
+                    if dependency in discovery:
+                        # Are the dependencies compiled?
+                        if discovery[dependency] == "Compiled":
+                            compiled[idep] = True
+                    else:
+                        # If the dependency isn't in the list at all yet
+                        # then add an entry so the system knows we are
+                        # expecting it later (for the above check)
+                        new_discovery[dependency] = "HeardOf"
```
MatthewHambley, line 129: This "discovery" dictionary may become an object in its own right with methods for updating the state of each unit. Another thought for the future.

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+                        new_discovery[definition] = "Compiled"
+                else:
+                    # If the dependencies weren't all satisfied then
+                    # back on the queue for another pass later
+                    new_artifacts.append(artifact)
+            else:
+                # If it wasn't required it could be later, so
+                # put it back on the queue, unless the target
+                # has been compiled, in which case it wasn't
+                # needed at all!
```
MatthewHambley, line 147: The comment talks about an error condition but the code doesn't seem to handle it. Maybe a ToDo?

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+                if discovery[self._target] != "Compiled":
+                    new_artifacts.append(artifact)
+
+        elif artifact.state is Compiled:
+            # Begin populating the list for linking
+            new_objects.append(artifact)
+            # But do not return a new artifact - this object
+            # is "done" as far as the processing is concerned
```
MatthewHambley, line 155: Why do we put the object back on the queue if it's done? Surely it should be dropped on the floor once it's compiled?

source/fab/engine.py
```diff
@@ -0,0 +1,190 @@
...
+            # by this point; so we can do the linking step
+            if self._target in artifact.defines:
+                task = self._taskmap[(artifact.filetype,
+                                      artifact.state)]
+                new_artifacts.extend(task.run(objects + [artifact]))
+
+        elif artifact.state is Linked:
+            # Nothing to do at present with the final linked
+            # executable, but included here for completeness
+            pass
```
MatthewHambley, line 168: In the future we might want to run it. e.g. for unit testing.

source/fab/queue.py
```diff
@@ -7,49 +7,66 @@
...
+    def __init__(self):
+        pass
 
-    @property
-    def products(self):
-        return []
 
-
-def worker(queue: JoinableQueue):
+def worker(queue: JoinableQueue,
```
MatthewHambley, line 25: Should probably be nominally private, i.e. `_worker(...`

source/fab/source_tree.py
```diff
@@ -6,80 +6,28 @@
...
-        elif issubclass(task_class, SingleFileCommand):
-            flags = self._command_flags_map.get(task_class, [])
-            task = CommandTask(
-                task_class(Path(reader.filename), self._workspace, flags))
-        else:
-            message = \
-                f"Unhandled class '{task_class}' in extension map."
-            raise TypeError(message)
-
-        self._task_handler(task)
```
MatthewHambley, line 79: Happy with the migration of this logic to the engine.

source/fab/tasks/fortran.py
```diff
@@ -303,9 +310,9 @@ def line_by_line(self) -> Iterator[str]:
             self._line_buffer = ''
 
 
-class FortranAnalyser(Analyser):
-    def __init__(self, reader: TextReader, database: SqliteStateDatabase):
-        super().__init__(reader, database)
+class FortranAnalyser(Task):
+    def __init__(self, workspace: Path):
+        self.database = SqliteStateDatabase(workspace)
```
MatthewHambley, line 315: Workspace is only used for instantiating the database. If the level above already has the database why not pass that instead?

source/fab/tasks/fortran.py
```diff
@@ -359,13 +366,26 @@ def __init__(self, reader: TextReader, database: SqliteStateDatabase):
...
         logger = logging.getLogger(__name__)
 
+        if len(artifacts) == 1:
+            artifact = artifacts[0]
+        else:
+            msg = ('Fortran Analyser expects only one Artifact, '
+                   f'but was given {len(artifacts)}')
+            raise TaskException(msg)
+
+        reader = FileTextReader(artifact.location)
```
MatthewHambley, line 379: Artefacts may gain the ability to return a reader object directly, in the future.

source/fab/tasks/fortran.py
```diff
@@ -464,80 +486,110 @@ def run(self):
...
+            msg = ('Fortran Preprocessor expects only one Artifact, '
+                   f'but was given {len(artifacts)}')
+            raise TaskException(msg)
 
-    @property
-    def as_list(self) -> List[str]:
-        base_command = ['cpp', '-traditional-cpp', '-P']
-        file_args = [str(self._filename), str(self.output[0])]
-        return base_command + self._flags + file_args
+        artifact = artifacts[0]
```
MatthewHambley, line 510: Hasn't this already been done in line 504?

source/fab/tasks/fortran.py
```diff
@@ -464,80 +486,110 @@ def run(self):
...
+        self._workspace = workspace
+
+    def run(self, artifacts: List[Artifact]) -> List[Artifact]:
+
+        if len(artifacts) == 1:
+            artifact = artifacts[0]
+        else:
+            msg = ('Fortran Compiler expects only one Artifact, '
+                   f'but was given {len(artifacts)}')
+            raise TaskException(msg)
```
MatthewHambley, line 543: Given that all these tasks are doing this check there may be space to hoist the functionality into a an abstract parent.

unit-tests/tasks/fortran_test.py
```diff
@@ -607,46 +658,3 @@ def test_preprocssor_output(self, caplog, tmp_path):
...
-        with pytest.raises(TaskException):
-            _ = test_unit.as_list
-
-        test_unit.add_object(Path('foo.o'))
-        assert test_unit.as_list == ['gfortran', '-o', 'bin/dusty',
-                                     'foo.o']
-
-        test_unit.add_object(Path('bar.o'))
-        assert test_unit.as_list == ['gfortran', '-o', 'bin/dusty',
-                                     'foo.o', 'bar.o']
```
MatthewHambley, line 652: Are we no longer testing these things?

source/fab/artifact.py
stevewardle: Hopefully it comes in useful - it's not totally clear to me yet exactly which files we will require hashes of, but now they'll just be available where needed

source/fab/builder.py
stevewardle: It seems like something which will make sense to do when we add C source (rather than writing a CLinker class that is almost identical, and because hopefully the linking stage is designed purely around linking binary objects and won't care or know what the language underneath is)

source/fab/builder.py
stevewardle: Do you think it's still worth having the callback (now that the callback only passes its arguments straight on)?  I deliberated over passing the queue instead and calling it directly but decided to keep it on the offchance we want to add some sort of additional processing to it later

source/fab/engine.py
stevewardle: I didn't want to raise an issue/ticket for something this minor so I've done it

source/fab/engine.py
stevewardle: Agreed - I guess it can probably do something be turned into something like a `List[TaskMap]` in a similar way to the `pathmaps` above it

source/fab/engine.py
stevewardle: Ah, of course!  It was easy to change the method to `__contains__` and then we can happily have `if artifact.location in pathmap` which is much nicer - thanks :+1:

source/fab/engine.py
stevewardle: It's definitely ripe for splitting up a bit - if `pylint` were here I expect it would be very upset.  I was thinking _maybe_ rather than a dictionary it may be possible to just use specific names - the branches which are more involved are _mostly_ describing actions which I hope are fairly language/configuration agnostic (e.g. one always has to do some sort of walk of the analysed objects and track compiles).  But yes either way let's not worry about it right now!

source/fab/engine.py
stevewardle: Perhaps poor wording of the comment - `artifact.state` is only accessible to the process which is currently looking at it - whereas here I was meaning the `discovery` array which is what allows _other processes_ to know that the units this artifact contains have been analysed.  I'll try to rephrase (or am I misunderstanding?)

source/fab/engine.py
stevewardle: I played about with this a little bit - though I may be wrong about this I couldn't quite see how to get the `Manager` (from `multiprocessing`) to let me use a generic object for these shared/proxy entities.  I agree it would be nicer if we can work it out though

source/fab/engine.py
stevewardle: I'm not sure it is an error condition.  It's the case where some of the Analysed files might not be needed to compile the target (this comes up in the preprocessing source test; depending on the macro you only need one of "stay" or "go now").  So this is just cycling the Analysed artifacts back onto the queue until they are either handled above (because their dependencies become Compiled) or the target gets compiled without them (so they weren't needed, and are dropped from the queue to let it finish)

source/fab/engine.py
stevewardle: I'm unsure what you mean - the objects don't go back onto the queue - `new_objects` is used to accumulate the shared `objects` array (at the end of the method) and the `artifact` is not put back on the queue.  The `objects` array is eventually used to provide the list of objects to the linking step (just below)

source/fab/engine.py
stevewardle: I thought _perhaps_ we might find some other use/s for it later - there could be a variety of things you might want to do as each target is produced (once we have multiple targets)

source/fab/tasks/fortran.py
stevewardle: I was thinking that eventually we won't want to have the block in `builder.py` which prints out the database contents to `stdout` - and since that is all that the top level uses the database for at this point we might be able to remove that from the top level.  Happy to change it though, what do you think?

source/fab/tasks/fortran.py
stevewardle: That's a nice idea, similar to pushing the hashing inside the object

source/fab/tasks/fortran.py
stevewardle: So it has!  I forgot to take it out after adding the trapping above; thanks :+1:

source/fab/tasks/fortran.py
stevewardle: Can I make the `@abstracmethod` in the `Task` parent class for `run` have this logic and then get each class to call `artifact = super().run(artifacts)` or would that be mis-using the way abstract methods are meant to be used?  Either that or I could add a class-level attribute which stores something like `input_type = 'Single'` (or `'Multiple'`) and provide a method the classes call like `artifact = self.check_input(artifacts)`... though that feels a little but cumbersome?  (Did you have a specific way in mind?)

unit-tests/tasks/fortran_test.py
stevewardle: I think we ought to still test the classes, but the previous tests were only really confirming that the old `Command` objects would correctly format the command.  Under the new design I can't think of how you would test them (we obviously don't want to run a compile / link to do so) - maybe there's something clever that can be done with mock/dummy objects to override `subprocess.run` so that instead of running something it returns the command it would have run and we can check it?

source/fab/builder.py
MatthewHambley: I think we retain the callback for the moment.

source/fab/tasks/fortran.py
MatthewHambley: That sounds reasonable. Lets just keep an eye on it so we're not backing life complicated for ourselves. No further changes required for the moment.

source/fab/tasks/fortran.py
MatthewHambley: Calling your parent implementation is absolutely a normal part of OO coding. The child implementation is modifying the behaviour of its parent in some way.

unit-tests/tasks/fortran_test.py
MatthewHambley: Given that this stuff is now happening in the engine it may fall under the testing for the engine.

unit-tests/artifact_test.py
```diff
@@ -0,0 +1,52 @@
...
+        assert artifact._hash is None
+
+    def test_hash(self, tmp_path: Path):
+        test_path = Path(tmp_path / 'test.foo')
+        test_path.write_text("Lorem ipsum dolor sit")
+        expected_hash = 1463158782
+        artifact = Artifact(test_path,
+                            Unknown,
+                            New)
+        assert artifact._hash is None
```
MatthewHambley, line 32: Not sure about testing nominally internal private variables. If someone goes accessing them then all bets are off. Therefore what value they contain is not important, or defined.

unit-tests/artifact_test.py
```diff
@@ -0,0 +1,52 @@
...
+        assert artifact.hash == expected_hash
+        # Check that it is cached
+        test_path.unlink()
+        assert artifact.hash == expected_hash
```
MatthewHambley, line 36: You raise an interesting question concerning how the artefact class should respond to the file it represents being removed from under it. I think the current behaviour is acceptable so no change needed.

unit-tests/artifact_test.py
stevewardle: I agree, _in principle_ - though I wanted to try and ensure that the hash hasn't been calculated prior to being requested (one of the problems with it being cached; checking it it is set via the `hash` property will cause it to be calculated!)  I can remove this if you think it's okay without it but can you think of any other way to check the hash part?

unit-tests/artifact_test.py
stevewardle: For what it's worth - I don't think we need to handle that situation; if we can't rely on files staying where we put (or find) them then there's a whole host of other things which could cause problems.  We can worry about this later but similar to the other comment above - I couldn't think of another way of checking that the hash is really cached

unit-tests/artifact_test.py
stevewardle: Thinking about it a little further - perhaps I can do something similar to the other test you commented on below - if I setup the artifact with a non-existent path then I _wouldn't_ expect it to raise whatever type of error it would raise when calling the constructor, but I _would_ expect it to get raised when I try to access the `hash` property (I'm not sure how sensible an idea this is though)

MatthewHambley: I think we now have 3 possible ways forward. I am happy to accept any one of them:

  * Leave it as it is
  * Remove the test on internal variables
  * Test on a none existent file.

stevewardle: Let's remove it I think - there may be a better way to test this which we can implement later

PR #83, opened 2020-07-30
Config implimentation

### LINKED ISSUE #68: Choose a configuration format and the tools to read it

GiorgiaLine:
This change impliments a simple .ini configuration file and parser.

Closes #68

### DIFF
--- a/source/fab/config.ini
+++ b/source/fab/config.ini
@@ -1,8 +1,0 @@
-[settings]
-target =
-exec-name =
-
-[flags]
-fpp-flags =
-fc-flags =
-ld-flags =
--- a/source/fab/config.ini
+++ b/source/fab/config.ini
@@ -1,0 +1,8 @@
+[settings]
+target =
+exec-name =
+
+[flags]
+fpp-flags =
+fc-flags =
+ld-flags =
--- a/system-tests/test.py
+++ b/system-tests/test.py
@@ -189,23 +189,23 @@ class RunFab(EnterPython):
         args.append(str(test_directory))
         args.append(str(conf_file))
 
-        configfile = open(conf_file, 'wt')
-        configfile.write('[settings] \n')
-        configfile.write('target = {}\n'.format(target))
-        configfile.write('exec-name = fab_test \n')
-        configfile.write('[flags] \n')
-        if fpp_flags:
-            configfile.write('fpp-flags = {}\n'.format(fpp_flags))
-        else:
-            configfile.write('fpp-flags = ' + '\n')
-        if fc_flags:
-            configfile.write('fc-flags = {}\n'.format(fc_flags))
-        else:
-            configfile.write('fc-flags = ' + '\n')
-        if ld_flags:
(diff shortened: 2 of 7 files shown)


system-tests/test.py
```diff
@@ -179,16 +180,37 @@ def __init__(self,
...
+        if fc_flags:
+            flags['fc-flags'] = fc_flags
+        else:
+            flags['fc-flags'] = ''
+        if ld_flags:
+            flags['ld-flags'] = ld_flags
+        else:
+            flags['ld-flags'] = ''
+        with open(conf_file, 'w') as configfile:
+            config.write(configfile)
```
MatthewHambley, line 213: Using the `configparser` library to build the test configuration file like this, while the correct course of action for application code, is probably not the best approach for test code. Tests tend to use static canned data wherever possible to avoid expanding the test surface. Every library included is effectively another moving part, another thing being tested.

In this case I would recommend a simple string and `.format()`.

source/fab/config.ini
```diff
@@ -0,0 +1,8 @@
+[settings]
+target =
+exec-name =
+
+[flags]
+fpp-flags =
+fc-flags =
+ld-flags =
```
MatthewHambley, line 8: Is this file left over from debug?

system-tests/test.py
GiorgiaLine: I can see your point and have changed how the test configuration is built.

source/fab/config.ini
GiorgiaLine: Yes it was. However now I'm wondering, related to your later question about a zero configuration mode of operation, if thee should be a default configuration file here? Or if I should have the configparser library build one from default values if one isn't present.

MatthewHambley: I think there are two modes of operation for zero configuration mode. The first is to simply build what is possible by making assumptions regarding what is to be built and how to do it. The second is to generate as much configuration file as possible using the same heuristics as the basis for a configuration driven build.

We probably don't want to get stuck in the weeds of trying to support either at this point. Just bare in mind that they will exist in the future. So lets remove it to keep this change clean.

system-tests/test.py
MatthewHambley: I can't see the change. Did you forget to "push" it?

GiorgiaLine: That's strange. I definitely pushed it, and can see it here on my end?

system-tests/test.py
```diff
@@ -179,16 +179,33 @@ def __init__(self,
...
+            conf_file = test_directory/('stay_config.ini')
+        else:
+            conf_file = test_directory/'config.ini'
 
-        args.extend(['--exec-name', 'fab_test'])
-        args.append(target)
         args.append(str(test_directory))
+        args.append(str(conf_file))
+
+        configfile = open(conf_file, 'wt')
```
MatthewHambley, line 192: I would recommend using the `with open(conf_file, 'wt') as configfile:` syntax here as it automatically handles a lot of fault conditions. You might also consider a """block string""" to save some typing.

PR #84, opened 2020-08-06
Sometimes the Subversion server test will fail

MatthewHambley:
This seems to be down to latency in the server opening its network connection. This fix adds a crude retry loop. It runs the risk of missing legitimate failures but should prevent spurious failures.

### DIFF
--- a/unit-tests/repository_test.py
+++ b/unit-tests/repository_test.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from subprocess import run, Popen
 import shutil
 import signal
+import time
 from typing import List, Tuple
 
 from pytest import fixture, mark, raises  # type: ignore
@@ -87,6 +88,7 @@ class TestSubversion:
             except FabException as ex:
                 if range == 0:
                     raise ex
+                time.sleep(1.0)
             else:
                 break
         _tree_compare(repo[1], tmp_path)
--- a/unit-tests/repository_test.py
+++ b/unit-tests/repository_test.py
@@ -81,7 +81,7 @@ class TestSubversion:
         # TODO: Is there a better solution such that we don't try to connect
         #       until the socket is open?
         #
-        for retry in range(3, 0 ,-1):
+        for retry in range(3, 0, -1):
             try:
                 test_unit.extract(tmp_path)
             except FabException as ex:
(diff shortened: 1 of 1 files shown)


MatthewHambley:
I've resubmitted the run a number of times and seen no failures. But then I never have before either.

unit-tests/repository_test.py
```diff
@@ -74,7 +74,21 @@ def test_extract_from_svn(self, repo: Tuple[Path, Path], tmp_path: Path):
...
+        #
+        # TODO: Is there a better solution such that we don't try to connect
+        #       until the socket is open?
+        #
+        for retry in range(3, 0, -1):
+            try:
+                test_unit.extract(tmp_path)
+            except FabException as ex:
+                if range == 0:
+                    raise ex
```
stevewardle, line 89: I think if you are re-raising the original exception you might be able to leave out the exception and just put `raise` - but at the same time it doesn't hurt anything to be explicit

unit-tests/repository_test.py
```diff
@@ -74,7 +74,21 @@ def test_extract_from_svn(self, repo: Tuple[Path, Path], tmp_path: Path):
         process = Popen(command)
 
         test_unit = SubversionRepo('svn://localhost/trunk')
-        test_unit.extract(tmp_path)
+        #
+        # It seems there can be a delay between the server starting and the
+        # listen socket opening. Thus we have a number of retries.
+        #
+        # TODO: Is there a better solution such that we don't try to connect
+        #       until the socket is open?
```
stevewardle, line 82: I had a read through the SVN documentation and can't see anything obvious

unit-tests/repository_test.py
```diff
@@ -74,7 +74,21 @@ def test_extract_from_svn(self, repo: Tuple[Path, Path], tmp_path: Path):
...
+        #
+        for retry in range(3, 0, -1):
+            try:
+                test_unit.extract(tmp_path)
+            except FabException as ex:
+                if range == 0:
```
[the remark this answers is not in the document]
MatthewHambley: A good point, busy waiting is rarely a good idea. I've added a 1s delay on failure.

unit-tests/repository_test.py
MatthewHambley: It's the classic problem with running it as a daemon. By definition it disconnects itself from the terminal and returns control to the launching process before it is finished setting up.

PR #86, opened 2020-08-14
Added a cleanup for the created config files

GiorgiaLine:
This change adds a teardown function to RunFab to clean up the config files created during testing.


system-tests/test.py
```diff
@@ -206,7 +206,7 @@ def __init__(self,
                 configfile.write('ld-flags = {}\n'.format(ld_flags))
             else:
                 configfile.write('ld-flags = ' + '\n')
-
+        self.conf_file = conf_file
```
stevewardle, line 209: Could replace `conf_file` with `self.conf_file` the whole way through?  Not a major issue but it would save having an extra intermediate variable

system-tests/test.py
```diff
@@ -219,6 +219,12 @@ def set_up(self):
         if self.test_parameters.work_directory.is_dir():
             shutil.rmtree(self.test_parameters.work_directory)
 
+    def tear_down(self):
+        """
+        Clean up config files following the run.
+        """
+        os.remove(self.conf_file)
```
stevewardle, line 226: Since `self.conf_file` is a `Path` object I think you could use one of its methods here?  `self.conf_file.unlink()` maybe?

stevewardle (approved):
Looks good to me

PR #90, opened 2021-04-16
Retire systest in favour of pytest

stevewardle:
We have been discussing and thinking for a while that we might want to drop `systest` - here we replace it in favour of using `pytest` to run both Unit and System testing.  Unsure if it's quite ready - this is a pretty light reworking just to prove that it runs; I suspect with a little more thought there may be a nicer design for it.  But it opens the discussion on how it should look at least

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,6 @@ setuptools.setup(
     extras_require={
         'dev': ['flake8', 'mypy'],
         'unit-test': ['pytest', 'pytest-cov', 'pytest-mock'],
-        'system-test': ['systest >= 5.5.0']
+        'system-test': ['pytest', 'pytest-cov', 'pytest-mock']
     }
 )
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,6 @@ setuptools.setup(
     extras_require={
         'dev': ['flake8', 'mypy'],
         'unit-test': ['pytest', 'pytest-cov', 'pytest-mock'],
-        'system-test': ['pytest', 'pytest-cov', 'pytest-mock']
+        'system-test': ['pytest']
     }
 )
--- a/system-tests/CUserHeader_test.py
+++ b/system-tests/CUserHeader_test.py
@@ -6,7 +6,7 @@
 from pathlib import Path
 from common import CompareConsoleWithFile, RunFab, RunExec, RunDump
 
-TEST_PATH = Path('system-tests/CUserHeader')
+TEST_PATH = Path('system-tests') / Path(__file__).name.split('_test.py')[0]
 
 
 def test_fab():
--- a/system-tests/CUserHeader_test.py
+++ b/system-tests/CUserHeader_test.py
@@ -1,0 +1,27 @@
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+from pathlib import Path
+from common import CompareConsoleWithFile, RunFab, RunExec, RunDump
+
+TEST_PATH = Path('system-tests/CUserHeader')
+
+
+def test_fab():
+    command = RunFab(TEST_PATH, 'main')
+    comparison = CompareConsoleWithFile(command)
+    comparison.run()
+
+
+def test_exec():
+    command = RunExec(TEST_PATH)
+    comparison = CompareConsoleWithFile(command)
+    comparison.run()
+
+
+def test_dump():
+    command = RunDump(TEST_PATH)
+    comparison = CompareConsoleWithFile(command)
+    comparison.run()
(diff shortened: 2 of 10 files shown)


setup.py
```diff
@@ -54,6 +54,6 @@
     extras_require={
         'dev': ['flake8', 'mypy'],
         'unit-test': ['pytest', 'pytest-cov', 'pytest-mock'],
-        'system-test': ['systest >= 5.5.0']
+        'system-test': ['pytest', 'pytest-cov', 'pytest-mock']
```
MatthewHambley, line 57: We probably shouldn't include the ability to "mock" in the system tests. If you are mocking you've probably gone wrong somewhere. System tests should be testing the whole system, not stubbing bits out.

system-tests/common.py
```diff
@@ -316,28 +309,27 @@ def tear_down(self):
             shutil.rmtree(self._repo_path)
 
 
-class CheckTask(systest.TestCase, metaclass=ABCMeta):
+class CheckTask(object, metaclass=ABCMeta):
```
MatthewHambley, line 312: I don't think it's necessary to explicitly inherit from `object` but it does no harm.

There is now a convenience class `ABC` you can inherit from. Check the documentation.

system-tests/common.py
```diff
@@ -352,9 +344,6 @@ class CompareConsoleWithFile(CheckTask):
     """
     # The expected result is held in a file "expected.<tag>[.<suffix>].txt.
     # Where "tag" comes from the task and "suffix" is specified.
-    #
-    # This comment is here because systest highjacks the doc string for output.
-    #
```
MatthewHambley, line 357: The point being that the comment can now once again go in the doc-string where it was supposed to be.

MatthewHambley (changes requested):
While there are potentially many improvements we could make to our testing I think this change should be kept simple. It achieves its objective which is to remove "systest".

Along with it we loose the stuff systest provided to run tests in parallel but mechanism by which that capability was provided lead to too much ugly code. Its absense certainly isn't a problem at the moment. If it becomes so later on we can reconsider then.

setup.py
stevewardle: Good point - I'll remove it.  While we're on the topic what about the coverage part?  Initially I was thinking it's useful to know how much of the codebase the system tests cover, but unless it's extremely clever I've a feeling it might not tell us much (seeing as most of the actual testing is happening inside `suprocess` calls to the various applications

system-tests/common.py
stevewardle: Ah right!  That makes sense, for some reason I thought the "This comment is here" only referred to itself and not the comment above, and I didn't spend long thinking about what it actually meant

stevewardle:
Should be ready for another look - as an aside I happened to spot that the coverage report wasn't running in the unit test output so I sorted that out too

system-tests/CUserHeader_test.py
```diff
@@ -0,0 +1,27 @@
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+from pathlib import Path
+from common import CompareConsoleWithFile, RunFab, RunExec, RunDump
+
+TEST_PATH = Path('system-tests/CUserHeader')
```
MatthewHambley, line 9: I'm a bit concerned that this `TEST_PATH` is appearing in every test package. Is there a way to improve this? Maybe their similarity illusory and they should be defined in each case.

system-tests/common.py
```diff
@@ -316,28 +309,27 @@ def tear_down(self):
             shutil.rmtree(self._repo_path)
 
 
-class CheckTask(systest.TestCase, metaclass=ABCMeta):
+class CheckTask(ABC, metaclass=ABCMeta):
```
MatthewHambley, line 312: Inheriting `ABC` implies a metalcass of `ABCmeta`.

MatthewHambley (changes requested):
I think we are neally ready to pull the trigger on this change.

system-tests/common.py
stevewardle: Ah yes, of course it does!

system-tests/CUserHeader_test.py
stevewardle: I wondered actually if it could be extracted from the routine name?  If we are going to keep a 1:1 correspondence with the test filename and the test directory name?

stevewardle:
How about this?  (though now a lot of the test cases seem so similar!  but I expect as we get further into having more functionality some future ones will be more complicated and justify having each with its own file)

MatthewHambley:
> How about this? (though now a lot of the test cases seem so similar! but I expect as we get further into having more functionality some future ones will be more complicated and justify having each with its own file)

Clearly there is still room for consolidation here but at least the duplicated code is clearly a candidate for removal.

It may be that what we are looking at here is a single test which takes the test name as a parameter. But that is outside the scope of this change. This change is just to make the move away from systest.

MatthewHambley (approved):
Let's light this candle.

PR #91, opened 2021-04-21
Allow debugging level to be specified and add more logging

stevewardle:
We have some existing debugging calls to the logger (in `fab.tasks.fortran.py`) but presently no way of specifying that logging level.  On the topic of logging also it would be helpful for other developments if there were slightly more logging on the other tasks so let's add a bit more to the debug level now that we can select it

A few thoughts I might appreciate the reviewer's input on:

 * I've copied what the `fortran` analyser does in the debug logging where it prints every line being "considered", only in the case of the `c` analyser it is iterating over `libclang` "node" objects - since this includes everything added by the preprocessor it ends up being a lot of spammy output.  I could restrict this so that it only prints interesting nodes to cut down on this (if we don't think that will lose too much helpfulness)
 * I contemplated moving the `getLogger` calls up to the module level, but hesitated in the end as I wasn't sure whether it was a good idea or not, but this does mean that they are duplicated a bit where files have multiple tasks
 * I updated the verbosity argument in all interfaces that had it despite it only really being utilised by the main `builder` one - figured it was sensible to be consistent, but maybe it's unnecessary?

### DIFF
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -62,8 +62,9 @@ def entry() -> None:
     parser.add_argument('-V', '--version', action='version',
                         version=fab.__version__,
                         help='Print version identifier and exit')
-    parser.add_argument('-v', '--verbose', action='store_true',
-                        help='Produce a running commentary on progress')
+    parser.add_argument('-v', '--verbose', action='count', default=0,
+                        help='Increase verbosity (may be specified once '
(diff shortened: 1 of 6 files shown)


GiorgiaLine (approved):
Looks to be ready for merging to me. A good start in adding propper loggong to fab.

PR #92, opened 2021-10-21
conda environment file

bblay:
Added conda environment file and readme. This is based on the [development environment wiki page](https://github.com/metomi/fab/wiki/Development-Environment). Using this I was able to run the unit tests with three skips.

Todo:
 - [ ] update wiki page if we merge this

### DIFF
--- a/Experimental/conda/environment.yml
+++ b/Experimental/conda/environment.yml
@@ -1,0 +1,13 @@
+name: fab
+dependencies:
+  - python=3.8
+  - pip=21.2.4
+  - flake8
+  - pytest
+  - pytest-mock
+  - pytest-cov
+  - libclang
+  - python-clang
+  - mypy
+  - pip:
+    - systest
--- a/Experimental/conda/environment.yml
+++ b/Experimental/conda/environment.yml
@@ -9,5 +9,3 @@ dependencies:
   - libclang
   - python-clang
   - mypy
-  - pip:
-    - systest
--- a/Experimental/conda/environment.yml
+++ b/Experimental/conda/environment.yml
@@ -1,4 +1,4 @@
-name: fab
+name: sci-fab
 dependencies:
   - python=3.8
   - pip=21.2.4
--- a/Experimental/conda/readme.md
+++ b/Experimental/conda/readme.md
@@ -1,0 +1,20 @@
+Create a conda environment for running fab
+```
+conda env create -f environment.yml
+```
+
+
+Activate the new environment
+
+```
+conda activate fab
+```
+
+Install fab (from the fab folder)
+```
+pip install .
+```
+
+Please be aware of some considerations when
+[using pip and conda](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#using-pip-in-an-environment)
+together.
--- a/Experimental/conda/readme.md
+++ b/Experimental/conda/readme.md
@@ -15,6 +15,11 @@ Install fab (from the fab folder)
 pip install .
 ```
 
+[Editable install](https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs) for developers
+```
+pip install -e .
+```
+
 Please be aware of some considerations when
 [using pip and conda](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#using-pip-in-an-environment)
 together.
--- a/Experimental/conda/readme.md
+++ b/Experimental/conda/readme.md
@@ -12,13 +12,25 @@ conda activate fab
 
 Install fab (from the fab folder)
 ```
+python setup.py install
+or
 pip install .
 ```
 
 [Editable install](https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs) for developers
 ```
+python setup.py develop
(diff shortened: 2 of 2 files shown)


wilbertcs (dismissed):

Experimental/conda/environment.yml
```diff
@@ -0,0 +1,11 @@
+name: fab
+dependencies:
+  - python=3.8
```
stevewardle, line 3: Is there a way in the syntax here to specify a minimum version instead of fixing to `3.8`?  (and if there is, would that allow someone to use this environment/recipe to build against a Python version of their choice provided it is recent enough?)

Experimental/conda/environment.yml
```diff
@@ -0,0 +1,11 @@
...
+  - pip=21.2.4
+  - flake8
+  - pytest
+  - pytest-mock
+  - pytest-cov
+  - libclang
+  - python-clang
+  - mypy
```
stevewardle, line 11: I wonder whether we should be explicit about the versions of everything?  (or set minimum versions as in the other comment).  I guess specifying the list without qualifiers just means get the most recent version of each package?

Experimental/conda/environment.yml
```diff
@@ -0,0 +1,11 @@
+name: fab
```
stevewardle, line 1: I may be misremembering but I think in other places we might have used the name sci-fab or something like that because the name `fab` was already taken on `conda-forge` - I'm just trying to remember _where_ we did this!

Experimental/conda/readme.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+[Editable install](https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs) for developers
+```
+pip install -e .
+```
+
+Please be aware of some considerations when
+[using pip and conda](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#using-pip-in-an-environment)
+together.
```
stevewardle, line 25: We could perhaps provide two alternate methods for the installation of `fab` itself?  I wasn't aware of the steps here and usually I do `python setup.py install` or `python setup.py develop` (for the equivalent of an "editable install") - So that would be "Install Fab (Pip)" or "Install Fab (setuptools)" or something like that?

Might also be worth a mention of how to un-install Fab from the environment (as could be needed if/when switching between working copies or similar) - for my method that would be `python setup.py install --uninstall`  (or `develop --uninstall`)

stevewardle (commented):
Looks fine and certainly the way we ought to be providing the details rather than having users copy paste a big line of setup from the wiki!  I guess longer term we might want to move this from the `Experimental` folder - is there a standard place this sort of thing tends to live in a Python project?

Experimental/conda/environment.yml
stevewardle: Found it - it's in the `setup.py`

Experimental/conda/environment.yml
bblay: Yes, the [docs seem to allow](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications) that. Not sure what our minimum spec is. I know Dave's got things running with 3.7 since his vdi is not playing ball, so we could use `>=3.7` or possible even just `=3`.

Experimental/conda/environment.yml
bblay: Yes, it'll get the latest version available to conda, including future bug fixes and potentially breaking changes for major versions. I don't know how we'd decide on minimum versions, other than pinning to whatever we're using right now, which seems arbitrary and might be newer than potential external collaborators have available. I suggest we leave it, especially as it's just copying the wiki command.

Experimental/conda/environment.yml
bblay: ooh, thank you

bblay: That's interesting, the name in setup.py doesn't match the github repo name - and we `import fab`, not sci-fab. I'll change the conda environment name to sci-fab but I think you've raised a new issue here - I think perhaps we should change the github repo name too.

Experimental/conda/environment.yml
bblay: Ah, however, the python and pip versions need to match...

Experimental/conda/environment.yml
bblay: I couldn't find a `fab` on conda-forge - not sure I searched in the right places?

bblay: Interestingly, to uninstall fab after installing with pip, i had to type `pip uninstall sci_fab`!

Experimental/conda/readme.md
```diff
@@ -0,0 +1,37 @@
+Create a conda environment for running fab
+```
+conda env create -f environment.yml
+```
+
+
+Activate the new environment
+
+```
+conda activate fab
```
rhattersley, line 10: This doesn't match the `sci-fab` value in the environment.yml file.

bblay: Thanks!

bblay:

PR #93, opened 2021-10-26
overview diagram

bblay:
This PR adds a simple diagram outlining the basic flow of Fab.

Overview diagram

Note: Previous diagrams-as-code have used [PlantUML](https://plantuml.com/), creating `.puml` files. For this PR I used the [diagrams.net](https://www.diagrams.net/) plugin for JetBrains/VS Code because the presentation quality is significantly higher. It produces SVG, XML or PNG.


bblay:
It's too ugly and I think there's a prettier option, which I'll spend 10 minutes investigating.

bblay:
Switching from plant uml to svg gives a much more professional look and diagrams.net (formerly draw.io) also has plugins for both VS and JetBrains. Also, github can render it. Hopefully this is a good way of having diagrams as code, going forward.

MatthewHambley (changes requested):
Not sure what your change is. SVG (Scalable Vector Graphics) is a file format, not a tool. Can you be more specific?

MatthewHambley:
Okay, but then this change is about chaging tools, not about changing files and it says nothing about tools.

bblay:
> Okay, but then this change is about chaging tools, not about changing files and it says nothing about tools.

You're right, I'll update the description.

bblay:

PR #96, opened 2021-10-27
Show procesing error message

bblay:
When a processing exception occurs, the worker will now display the error message and optionally try to exit the program.

The program doesn't actually exit, though. I think we might be seeing the [_Joining processes that use queues_](https://docs.python.org/3/library/multiprocessing.html#all-start-methods) problem? However, the user can now see that an error has occurred, with detail about the error, and be advised to exit manually.

Sample output with `stop_on_error=True`:
```
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/templating_mod.f90:
  Expected end of module "templating_mod" but found function: END FUNCTION tpl_has_var_name
Please exit with ctrl-c
```

Sample output with `stop_on_error=False`:
```
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/templating_mod.f90:
  Expected end of module "templating_mod" but found function: END FUNCTION tpl_has_var_name
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/string_utils_mod.f90:
  Expected end of if "None" but found function: END FUNCTION str_starts_with
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/metstats_timestep.f90:
  Expected end of if "None" but found end of where: END WHERE
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/metstats_mod.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE metstats_allocate
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/metstats_init.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE metstats_init
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/logging_mod.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE read_nml_jules_prnt_control
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/grid_utils_mod.f90:
  Expected end of if "None" but found function: END FUNCTION grid_create
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/dictionary_mod.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE dict_get_int
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/datetime_mod.f90:
  Expected end of module "datetime_mod" but found function: END FUNCTION datetime_clone
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/data_cube_mod.f90:
  Expected end of module "data_cube_mod" but found function: END FUNCTION is_mdi
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/cable_pack_mod.f90:
  Expected end of if "None" but found do:  END DO
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/water_resources_drive.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE water_resources_drive
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/vegcarb_jls.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE vegcarb
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/veg-veg2a_jls_mod.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE veg2
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/veg-veg1a_jls_mod.f90:
  Expected end of if "None" but found subroutine: END SUBROUTINE veg1
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/triffid_jls.f90:
  Expected end of if "None" but found do: END DO
ERROR processing /home/h02/bblay/git/fab/tmp-workspace/sparm_jls_mod.f90:
  Expected end of if "None" but found do:  END DO
  ...
  ```

### DIFF
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -72,6 +72,7 @@ def entry() -> None:
                         choices=range(2, multiprocessing.cpu_count()),
                         help='Provide number of processors available for use,'
                              'default is 2 if not set.')
+    parser.add_argument('--stop-on-error', default=True)
     parser.add_argument('source', type=Path,
                         help='The path of the source tree to build')
     parser.add_argument('conf_file', type=Path, default='config.ini',
@@ -98,7 +99,8 @@ def entry() -> None:
                       flags['fpp-flags'],
                       flags['fc-flags'],
                       flags['ld-flags'],
-                      arguments.nprocs)
+                      arguments.nprocs,
+                      arguments.stop_on_error)
     application.run(arguments.source)
 
 
@@ -110,7 +112,8 @@ class Fab(object):
                  fpp_flags: str,
                  fc_flags: str,
                  ld_flags: str,
-                 n_procs: int):
+                 n_procs: int,
+                 stop_on_error: bool=True):
 
         self._workspace = workspace
         if not workspace.exists():
@@ -177,7 +180,7 @@ class Fab(object):
                         target,
                         path_maps,
                         task_map)
-        self._queue = QueueManager(n_procs - 1, engine)
+        self._queue = QueueManager(n_procs - 1, engine, stop_on_error)
 
     def _extend_queue(self, artifact: Artifact) -> None:
         self._queue.add_to_queue(artifact)
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -72,7 +72,7 @@ def entry() -> None:
                         choices=range(2, multiprocessing.cpu_count()),
                         help='Provide number of processors available for use,'
                              'default is 2 if not set.')
-    parser.add_argument('--stop-on-error', default=True)
+    parser.add_argument('--stop-on-error', default=True, action="store_true")
     parser.add_argument('source', type=Path,
                         help='The path of the source tree to build')
     parser.add_argument('conf_file', type=Path, default='config.ini',
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -72,7 +72,7 @@ def entry() -> None:
                         choices=range(2, multiprocessing.cpu_count()),
                         help='Provide number of processors available for use,'
                              'default is 2 if not set.')
-    parser.add_argument('--stop-on-error', default=True, action="store_true")
+    parser.add_argument('--stop-on-error', action="store_true")
     parser.add_argument('source', type=Path,
                         help='The path of the source tree to build')
     parser.add_argument('conf_file', type=Path, default='config.ini',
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by wilbertcs — APPROVED** [review]

File: source/fab/builder.py [hunk 1]
```diff
@@ -72,6 +72,7 @@ def entry() -> None:
                         choices=range(2, multiprocessing.cpu_count()),
                         help='Provide number of processors available for use,'
                              'default is 2 if not set.')
+    parser.add_argument('--stop-on-error', action="store_true")
```
MatthewHambley, line 75: Will we ever need to not stop on an error? Errors tend to be pretty terminal.

source/fab/tasks/fortran.py
```diff
@@ -537,19 +537,21 @@ def run(self, artifacts: List[Artifact]) -> List[Artifact]:
                 if end_nature is not None:
                     if end_nature != exp[0]:
                         end_message = 'Expected end of {exp} "{name}" ' \
-                                      'but found {found}'
+                                      'but found {found}: {line}'
```
MatthewHambley, line 540: This seems unrelated to stopping on error but it's such a small thing I don't see a need to remove it.

MatthewHambley (changes requested):
Unless there's a good reason for it I think that user control of the stop behaviour can be removed.

source/fab/builder.py
bblay: It's definitely useful to see all the errors in one output stream rather than having to keep fixing one problem at a time then re-running. With the latter you don't get an idea of how far through the fixing process you are. With the former you can fix lots of stuff at once if you like. Better still, if you can see all the classes of errors that you need to fix, you can potentially pick a better solution sooner.

source/fab/tasks/fortran.py
bblay: Agreed, this PR about error messages.

source/fab/builder.py
MatthewHambley: In which case would you ever want to stop on an error? Would you not prefer to view all errors as you suggest?

bblay: Good question. I've been using it. It's useful to stop on the first error sometimes so you don't have to search the output for the first one. This flag seems likely to evolve into the more sophisticated behaviour the team's been discussing. I think this is a reasonable step in the right direction.

bblay: I'd somehow missed the request to remove this from _user_ control, now done.

bblay:

PR #97, opened 2021-10-28
handle "logical function"

bblay:
Handle functions preceded with the word "Logical".
Authored with Dave.


wilbertcs (approved):

bblay:
Having realised the "logical" was the boolean return type, we're planning a broader solution to this bug which caters for lots of different return types, including custom types.

bblay:
I should probably reopen this as subsequent improvements might be shelved, so this would represent how far we got.

PR #99, opened 2021-11-11
Fparser2

bblay:
This PR replaces the in-house Fortran parser with fparser2.

It's also built upon a few other open PRs, which should probably be merged or closed first.

### DIFF
--- a/Experimental/conda/environment.yml
+++ b/Experimental/conda/environment.yml
@@ -1,0 +1,13 @@
+name: sci-fab
+dependencies:
+  - flake8
+  - libclang
+  - mypy
+  - python=3.8
+  - pip=21.2.4
+  - pytest
+  - python-clang
+  - pytest-cov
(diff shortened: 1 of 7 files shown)


bblay:
rebuild

bblay:
Ah, I've used walrus operator for the first time because they are nice and tidy, but that requires Python 3.8. Is it ok to bump setup.py to 3.8? If not, I can just replace the walrus operators instead, no problem.

stevewardle:
> Ah, I've used walrus operator for the first time because they are nice and tidy, but that requires Python 3.8. Is it ok to bump setup.py to 3.8? If not, I can just replace the walrus operators instead, no problem.

I think I'd avoid restricting us to 3.8+ just for this feature

bblay:
I'm going to close this as we're really just working on one big feature branch.

PR #103, opened 2022-04-06
fix logging

bblay:
During a recent review we broke some of the logging!


GiorgiaLine (approved):
Looks like a completely sensible change. Approved.

bblay:
Thanks Giorgia!

PR #104, opened 2022-04-07
run_command capture stderr

bblay:
`run_command`now captures stderr and includes it in it's raised exception


stevewardle (approved):
Seems all good to me

PR #106, opened 2022-04-19
Metrics

bblay:
Adds metrics, including some per-artefact metrics and some matplotlib charts.
pie
compile_fortran

### DIFF
--- a/source/fab/builder.py
+++ b/source/fab/builder.py
@@ -8,14 +8,18 @@ A build config contains a list of steps, each with a run method which it calls o
 Each step can access the artifacts created by previous steps, and add their own.
 
 """
+import getpass
 import logging
 import multiprocessing
+import os
 from datetime import datetime
 from pathlib import Path
 
+from fab.metrics import init_metrics, stop_metrics, metrics_summary, send_metric
+
 from fab.config import Config
 from fab.constants import BUILD_OUTPUT
-from fab.util import time_logger
+from fab.util import TimerLogger
 
 logger = logging.getLogger(__name__)
 
@@ -67,6 +71,7 @@ def entry() -> None:
 class Build(object):
     def __init__(self, config: Config):
         self.config = config
+        self.metrics_folder = None
 
         if not config.workspace.exists():
             config.workspace.mkdir(parents=True)
@@ -74,12 +79,48 @@ class Build(object):
             (config.workspace / BUILD_OUTPUT).mkdir()
 
     def run(self):
-        logger.info(f"{datetime.now()}")
-        logger.info(f"use_multiprocessing = {self.config.use_multiprocessing}")
-        if self.config.use_multiprocessing:
-            logger.info(f"n_procs = {self.config.n_procs}")
+        self.init_logging()
+        self.init_metrics()
 
         artefact_store = dict()
-        for step in self.config.steps:
-            with time_logger(step.name):
-                step.run(artefact_store, self.config)
+        try:
+            with TimerLogger(f'running {self.config.label} build steps') as steps_timer:
+                for step in self.config.steps:
+                    with TimerLogger(step.name) as step_timer:
+                        step.run(artefact_store=artefact_store, config=self.config)
+                    send_metric('steps', step.name, step_timer.taken)
+        except Exception as err:
+            raise Exception(f'\n\nError running build steps:\n{err}')
+        finally:
+            self.finalise_metrics(steps_timer.taken)
+
+    def init_logging(self):
+        logger.info(f"{datetime.now()}")
+        if self.config.use_multiprocessing:
+            logger.info(f'machine cores: {multiprocessing.cpu_count()}')
+            logger.info(f'available cores: {len(os.sched_getaffinity(0))}')
+            logger.info(f'using n_procs = {self.config.n_procs}')
+        logger.info(f"workspace is {self.config.workspace}")
+
+    def init_metrics(self):
+        self.metrics_folder = self.config.workspace / 'metrics' / self.config.label.replace(' ', '_')
+
+        init_metrics(metrics_folder=self.metrics_folder)
+
+        send_metric(group='run', name='label', value=self.config.label)
+        send_metric(group='run', name='datetime', value=datetime.now().replace(microsecond=0).isoformat())
+        send_metric(group='run', name='sysname', value=os.uname().sysname)
+        send_metric(group='run', name='nodename', value=os.uname().nodename)
+        send_metric(group='run', name='machine', value=os.uname().machine)
+        send_metric(group='run', name='user', value=getpass.getuser())
+
+        if self.config.use_multiprocessing:
+            send_metric(group='run', name='machine cores', value=multiprocessing.cpu_count())
+            send_metric(group='run', name='available cores', value=len(os.sched_getaffinity(0)))
+            send_metric(group='run', name='using n_procs', value=self.config.n_procs)
+
+    def finalise_metrics(self, time_taken):
+        send_metric(group='run', name='time taken', value=time_taken)
+
+        stop_metrics()
+        metrics_summary(metrics_folder=self.metrics_folder)
--- a/source/fab/metrics.py
+++ b/source/fab/metrics.py
@@ -1,0 +1,201 @@
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+"""
+A module for recording and summarising metrics, with the following concepts:
+
+init
+    create pipes
+    create reading process - daemonic, exits with main process
+
+send
+    group, name, value -> reading process
+    overwrites any previous value for group[name]
+
+reading process
+    creates and add to metrics dict
+    finishes -> send whole lot down a summary pipe and close
+
+stop
+    closes pipes & process
+    return metrics from summary pipe
+
+"""
+import datetime
+import json
+import logging
+from collections import defaultdict
+from multiprocessing import Process, Pipe
+from multiprocessing.connection import Connection
+from pathlib import Path
+from typing import Optional, Dict
+
+JSON_FILENAME = 'metrics.json'
+
+logger = logging.getLogger('fab')
+
+# the pipe for individual metrics
+_metric_recv_conn: Optional[Connection] = None
+_metric_send_conn: Optional[Connection] = None
+
+# the process which receives individual metrics
+_metric_recv_process: Optional[Process] = None
+
+
+def init_metrics(metrics_folder):
+    """
+    Create the pipe for sending metrics, the process to read them, and another pipe to push the final collated data.
+
+    Only one call to init_metrics can be called before calling stop_metrics.
+
+    """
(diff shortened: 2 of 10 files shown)


source/fab/builder.py
```diff
@@ -74,12 +78,48 @@ def __init__(self, config: Config):
...
+
+    def init_logging(self):
         logger.info(f"{datetime.now()}")
-        logger.info(f"use_multiprocessing = {self.config.use_multiprocessing}")
         if self.config.use_multiprocessing:
-            logger.info(f"n_procs = {self.config.n_procs}")
+            logger.info(f'machine cores: {multiprocessing.cpu_count()}')
+            logger.info(f'available cores: {len(os.sched_getaffinity(0))}')
+            logger.info(f'using n_procs = {self.config.n_procs}')
+        logger.info(f"workspace is {self.config.workspace}")
```
MatthewHambley, line 102: It would be well worth taking a few moments to nail down how you want to use the different logging levels in this project. We failed to do that for Gung Ho and now everything is either "info" or an error.

How much of this output do you want to see in a normal run and how much should be for debug purposes only?

source/fab/builder.py
```diff
@@ -74,12 +78,48 @@ def __init__(self, config: Config):
...
+        send_metric(group='run', name='datetime', value=datetime.now().replace(microsecond=0).isoformat())
+        send_metric(group='run', name='sysname', value=os.uname().sysname)
+        send_metric(group='run', name='nodename', value=os.uname().nodename)
+        send_metric(group='run', name='machine', value=os.uname().machine)
+        send_metric(group='run', name='user', value=getpass.getuser())
+
+        if self.config.use_multiprocessing:
+            send_metric(group='run', name='machine cores', value=multiprocessing.cpu_count())
+            send_metric(group='run', name='available cores', value=len(os.sched_getaffinity(0)))
+            send_metric(group='run', name='using n_procs', value=self.config.n_procs)
```
MatthewHambley, line 117: None of these seem like metrics. They seem more like technical diagnostics. Potentially their own logging stream. 

Just seeing the name "metrics" makes me think of time series data which none of this is.

Is this a problem with the name not reflecting the purpose or with overloading something intended for something else?

source/fab/metrics.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # start the receiving process
+    _metric_recv_process = Process(
+        target=_read_metric,
+        daemon=True,  # todo: test this thoroughly, manually
+    )
+    _metric_recv_process.start()
+
+
+def _read_metric():
```
MatthewHambley, line 77: Normally the pattern here would be to pass the necessary pipe endpoints as context to the process, rather than using globals. I think the `args` argument of the `Process` constructor allows for this.

source/fab/metrics.py
```diff
@@ -0,0 +1,214 @@
...
+    logger.debug('read_metric: waiting for metrics')
+    num_recorded = 0
+    while True:
+        try:
+            metric = _metric_recv_conn.recv()
+        except EOFError:
+            break
+
+        group, name, value = metric
+        metrics[group][name] = value
```
MatthewHambley, line 104: Obviously there is scope here to over run available memory during execution. Future development should look to buffering and spooling out blocks of records.

source/fab/metrics.py
```diff
@@ -0,0 +1,214 @@
...
+        num_recorded += 1
+
+    logger.debug(f"read_metric: recorded {num_recorded} metrics")
+
+    # Send the collated metrics.
+    # Note: We tried a 2-way pipe so we'd only need one, but it didn't work
+    # so we moved on and created this second pipe for the total metrics.
+    # Might be a simplification in there.
+    _collated_metrics_send_conn.send(metrics)
+    _collated_metrics_send_conn.close()
```
MatthewHambley, line 114: What is the reason for sending the collated records over a pipe? Why not write them out from this process?

source/fab/metrics.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # set these to none so metrics can be initialised again
+    _metric_recv_conn = _metric_send_conn = _metric_recv_process = _collated_metrics_recv_conn = None
+
+    return metrics
+
+
+def metrics_summary(metrics, output_folder: Path):
+    """
+    Create various summaries fo the metrics, including charts if matplotlib is installed.
```
MatthewHambley, line 153: Typo

source/fab/metrics.py
```diff
@@ -0,0 +1,214 @@
...
+        plt = None
+
+    logger.debug(f'metrics_summary: got metrics for: {metrics.keys()}')
+    output_folder.mkdir(parents=True, exist_ok=True)
+
+    metrics_folder = Path(output_folder)
+    metrics_folder.mkdir(exist_ok=True)
+
+    with open(metrics_folder / 'metrics.json', 'wt') as outfile:
+        json.dump(metrics, outfile, indent='\t')
```
MatthewHambley, line 170: Given that this is not Javascript and the ugliness factor, what is the reason for using JSON?

source/fab/metrics.py
```diff
@@ -0,0 +1,214 @@
...
+    # pull out the total metrics
+    metrics = _collated_metrics_recv_conn.recv()  # type: ignore
+
+    # set these to none so metrics can be initialised again
+    _metric_recv_conn = _metric_send_conn = _metric_recv_process = _collated_metrics_recv_conn = None
+
+    return metrics
+
+
+def metrics_summary(metrics, output_folder: Path):
```
MatthewHambley, line 151: For the future, consider refactoring this to be more modular making it easier to turn different plots (and data sets) on and off. Also to add new ones and remove old ones.

source/fab/steps/compile_fortran.py
```diff
@@ -113,21 +115,23 @@ def get_compile_next(self, already_compiled_files: Set[Path], to_compile: Set[An
...
+            command.append(str(analysed_file.fpath))
 
-        output_fpath = analysed_file.fpath.with_suffix('.o')
-        if self._config.debug_skip and output_fpath.exists():
-            log_or_dot(logger, f'Compiler skipping: {output_fpath}')
-            return CompiledFile(analysed_file, output_fpath)
+            output_fpath = analysed_file.fpath.with_suffix('.o')
+            if self._config.debug_skip and output_fpath.exists():
+                log_or_dot(logger, f'Compiler skipping: {output_fpath}')
+                return CompiledFile(analysed_file, output_fpath)
```
MatthewHambley, line 126: Does the timer work correctly with the premature return?

source/fab/util.py
```diff
@@ -50,12 +50,54 @@ def file_walk(path: Path) -> Iterator[Path]:
...
+
+    def __enter__(self):
+        self._start = perf_counter()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.taken = perf_counter() - self._start
+
+
+class TimerLogger(object):
```
MatthewHambley, line 70: Are both `Timer` and `TimerLogger` necessary or is one left over from earlier development? There seems to be duplicated functionality. Could they inherit?

MatthewHambley (changes requested):
Overall this looks fine. Just a request to consider if the design and terminology are as clear as they could be. Also please consider whether some or all of this functionality could be an extension of the Python logging framework. I believe it allows arbitrary handlers to be written. It may also support additional logging streams which could be used to manage "metrics".

bblay:
On the use of a logger for recording metrics: On one hand, it's good that it would remove the cost of maintaining the pipe and recorder code, etc. On the other hand, it seems we'd be forced to introduce the concept of log levels to our metrics recording, when those two concepts are unrelated. That's a code smell to me so I'd argue that logging is not a good fit in this case. Always happy to discuss this further, of course. There's a third option: use a third party library like [prometheus](https://github.com/prometheus/client_python). This would be a great solution if we're happy to introduce this new dependency.

source/fab/builder.py
bblay: I think that work was done by the wider community before us, and we can stand on their shoulders. The language gives us _debug,_ _info,_ _warning_ and _error_, and I think they're self-explanatory. So it's a debug message if you only want to see it when something's going wrong. It's an info message if you want to see light-weight, general progress through a run. A warning if you want a quieter run.

In our current mode of operation we like clean, concise, info-level logging so we can keep an eye on a run and look back when something's not right. We regularly tweak the logging to show only what's interesting to us. Some of these messages may move to debug in to future, and we might revert the logging level to warning but I guess it's too early to speculate on that.

source/fab/builder.py
bblay: Some metrics only become a time series over multiple runs. The classic example is the response time of a server's endpoint. You'd record that single metric every run and analyse the trend over time. This is where a tool like Prometheus could be useful as it can [do all that](https://prometheus.io/docs/visualization/grafana/) for us.

I think you're right to question whether the number of cores, etc, is useful in metrics. The reason they're there right now is merely for annotating the charts. Like logging, I expect we'll be tweaking what metrics are recorded as development progresses. In the future we might want to visualise how run time is affected by the number of utilised cores, which would presumably make this a useful metric but it seems too early to speculate. Always happy to chat further in case I've missed something.

source/fab/metrics.py
bblay: Agreed. We've raised a [ticket to consider metrics](https://github.com/metomi/fab/projects/8#card-81420757), which might make this redundant.

source/fab/metrics.py
bblay: Agreed (but unlikely?). Again, a third party tool would cover this so we'd like to postpone but we've also added a _todo_ in the code.

source/fab/metrics.py
bblay: Yes that's nicer, thank you.

source/fab/builder.py
MatthewHambley: I think you'd be amazed at how many different interpretations of the concept "info" there are but if you are happy with it as is I don't have a problem. As you say, it can all be reorganised later.

source/fab/builder.py
MatthewHambley: As long as there is a plan which includes reassessing current usage I am okay with it as is.

source/fab/metrics.py
MatthewHambley: I've updated the note with this point.

source/fab/metrics.py
bblay: :smiley: It's built into Python for a reason. This is a really easy way to save and load a dict with minimal code to maintain. Happy to consider alternatives, of course.

source/fab/steps/compile_fortran.py
bblay: Yep, _"The with statement guarantees that if the __enter__() method returns without an error, then __exit__() will always be called."_ [[docs]](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement)

MatthewHambley (dismissed):
This looks like a useful tool for understanding and tracking optimisations.

MatthewHambley (approved):
Everything is still okay.

PR #110, opened 2022-05-17
a few minor simplifications to gcom run configs

bblay:
We've shown Fab to Andy Malcolm and wanted to make a few small simplifications to make it as easy as possible for him.

### DIFF
--- a/run_configs/gcom/gcom_build_steps.py
+++ b/run_configs/gcom/gcom_build_steps.py
@@ -15,7 +15,7 @@ from fab.steps.walk_source import FindSourceFiles
 
 
 def common_build_steps(fpic=False) -> List[Step]:
-    steps: List[Step] = [
+    steps = [
         FindSourceFiles(),
         c_preprocessor(),
         fortran_preprocessor(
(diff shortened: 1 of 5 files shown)


run_configs/gcom/gcom_build_steps.py
```diff
@@ -15,7 +15,7 @@
 
 
 def common_build_steps(fpic=False) -> List[Step]:
-    steps: List[Step] = [
+    steps = [
```
MatthewHambley, line 18: Not sure how removing the type hinting makes it easier to run.

source/fab/metrics.py
```diff
@@ -160,7 +160,7 @@ def metrics_summary(metrics_folder: Path):
     step_names = ['preprocess fortran', 'preprocess c', 'compile fortran']
     for step_name in step_names:
         if step_name not in metrics:
-            logger.error(f'no step metrics called {step_name}')
+            # logger.error(f'no step metrics called {step_name}')
```
MatthewHambley, line 163: If code isn't needed then remove it. Otherwise it is needed and shouldn't be commented. Is the problem that the error message needs to be reworded for clarity?

MatthewHambley (changes requested):
Just a few monior issues.

run_configs/gcom/gcom_build_steps.py
bblay: It was intended to make it easier to read for someone who doesn't know Python at all. We wanted it super-lightweight.

MatthewHambley (approved):
As noted, I don't think it's unreasonable to expect science developers to know our standard scripting language however I don't intend to hold this change up on that.

PR #114, opened 2022-05-25
Misc tidy 1

bblay:
A few small tidy ups from a long-lived branch which we're pulling into master:
 - Moved `by_type()` to utils
 - Renamed a few more variables and params from _workspace_ to _project_workspace_. Some of these changes had already been brought into master from that branch, this completes it. Motivation: it was sometimes confusing/ambiguous as there's the both the _fab_ workspace and the _project_ workspace.
 - Step is now abstract, rather that just the base for grab steps, which is more correct.
 - The `FindSourceFiles` step no longer creates the output folders. That's not its job and folders are now created elsewhere now.
 - Added our helper script to run all build configs, as we currently do this frequently as a manual test.

### DIFF
--- a/run_configs/test_all.sh
+++ b/run_configs/test_all.sh
@@ -1,0 +1,12 @@
+set -e
+
+cd ~/git/fab/run_configs/gcom
+python -m grab_gcom
+python -m build_gcom_ar
+python -m build_gcom_so
+
+cd ../jules
+python -m build_jules
+
+cd ../um
+python -m build_um
--- a/source/fab/steps/grab.py
+++ b/source/fab/steps/grab.py
@@ -24,7 +24,7 @@ class GrabBase(Step, ABC):
             - dst_label: The name of a sub folder in the project workspace, in which to put the source.
 
         """
-        super().__init__(name=name or f'{self.__class__.__name__} {dst_label}'.strip())
+        super().__init__(name=name or f'{self.__class__.__name__} {dst_label or src}'.strip())
         self.src: str = src
         self.dst_label: str = dst_label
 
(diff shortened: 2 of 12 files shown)


run_configs/test_all.sh
```diff
@@ -0,0 +1,12 @@
...
+cd ~/git/fab/run_configs/gcom
+python -m grab_gcom
+python -m build_gcom_ar
+python -m build_gcom_so
+
+cd ../jules
+python -m build_jules
+
+cd ../um
+python -m build_um
```
MatthewHambley, line 12: I notice that Gung Ho is not part of "all." There appears to be a change ordering issue here.

MatthewHambley (changes requested):
Just a couple of things.

run_configs/test_all.sh
bblay: Yep, that's in a separate PR and we'll be adding it here when it's merged, sooner or later.

source/fab/steps/grab.py
```diff
@@ -4,15 +4,14 @@
...
-from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Dict
 
 from fab.steps import Step
 from fab.util import run_command
 
 
-class GrabBase(Step, ABC):
+class GrabBase(Step):
```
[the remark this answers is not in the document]
bblay: Ahh, good question. It _is_ still abstract. We should have made Step abstract _as well_, not instead! Will do that now, thank you.

MatthewHambley (approved):
This change is mostly about renaming things for clarity.

PR #115, opened 2022-05-25
Misc logic changes 1

bblay:
A few small logic changes from a long-lived branch that we're bringing into master.
 - Allow `FilterBuildTree` to accept a single suffix as a string.
 - Log file handler now attached to fab logger instead of root.
 - Simplify params to `add_mo_commented_file_deps`, including new helper `filter_source_tree`.


MatthewHambley (approved):
As advertised, just some minor tidying up.

PR #116, opened 2022-05-26
Remove endpoints

bblay:
A new user tried to use an old entry point. This PR removes them (for now at least) to avoid further confusion.


MatthewHambley (approved):
If these entry points don't work then we shouldn't be carrying them around.

PR #120, opened 2022-06-24
Get rose suites running

bblay:
Perhaps a couple of months ago, our conda env stopped working standalone and had to be used together with the lfric modules commands. We haven't updated the rose configs to use these modules until now. This PR does that, and introduces rose edit meta for version number selection in many of the rose suites.


GiorgiaLine (approved):
Everything looks in line with the rose config files seen in the UM and JULES, so I think this is good to go on.

PR #124, opened 2022-07-07
Build docs with github actions

bblay:
A manually triggered github actions workflow to build the docs from any branch, then push to the github_pages branch.


.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  build-docs:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: set git user
+      run: |
+        git config --global user.email "metomi@metoffice.gov.uk"
```
MatthewHambley, line 14: I don't think this e-mail address is monitored any more even if it exists. Rich may have an idea for a more suitable option.

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,42 @@
...
+        make html
+    - name: move built docs to docs root
+      run: |
+        mv docs/build/html/* docs/
+    - name: git add built docs
+      run: |
+        git add docs/*
+    - name: commit
+      run: |
+        git commit -m "docs build"
```
MatthewHambley, line 38: I'm always wary of committing generated files to a source repository. Is there a reason not to build and commit in the `gh_pages` branch?

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,42 @@
+# Please keep in sync with /environment.yml
```
MatthewHambley, line 1: What is it that has to be kept in sync? Is there a way to ensure that it is?

MatthewHambley (changes requested):
Seeking clarification on operation.

.github/workflows/build_docs.yml
bblay: This is the same as on PyPI, and Richard is now on holiday - sorry we left this so long! Perhaps we should make a ticket to cover both this and the PyPI email instead?

.github/workflows/build_docs.yml
bblay: The risk seems the same. Perhaps I'm missing something?
 - Checkout the given branch --> build --> force push to gh_pages
 - Checkout the given branch --> force replace gh_pages --> build --> push

.github/workflows/build_docs.yml
bblay: Oh that's a copy-paste relic from the main build action (where it refers to setting up the environment where we run Fab). Will remove.

.github/workflows/build_docs.yml
MatthewHambley: Done, see #130

.github/workflows/build_docs.yml
MatthewHambley: The way I did it with Stylist was to checkout both the project and the gh_pages. Then I could copy across. The whole gh_pages thing is rather inelligant whichever way it's done but it's what we have.

MatthewHambley (approved):
Since the best practice for publishing documentation is not clear we're going to try different things here than have already been done with Stylist.

PR #126, opened 2022-07-11
Docstrings sweep

bblay:


MatthewHambley (changes requested):
Adds an refines a lot of docstrings, as advertised. Also some minor code reorganisation and an argument rename.

For interest I tried running `pycodestyle` over the source directory and got a lot of lines over 79 characters long.

I also tried `pylint` which returned a 7.08 out of 10 score.

I'm not necessarily suggesting we use either of these tools but thought I would give them a whirl to see what they reported. Adopting either one would take some effort but might be worth it.

MatthewHambley (approved):
A great improvement in documentation.

PR #131, opened 2022-08-01
Incremental build: Fortran

bblay:
Fortran compiler now records the hash of the _source, flags, newly created modules and module dependencies_ at the time of compilation, and uses that information to decide if a file needs recompiling.

Deferring the planning and introduction of a new database technology, this PR uses csv like the analysis step.

Also in this PR, in order of appearance:
 - Some variable/constant name changes for readability.
 - Pull out the compile pass into it's own method, _compile_pass_.
 - Pull out the artefact recording into it's own method, _store_artefacts_.
 - Due to the extra logic needed in the subprocesses, _compile_file_ is now wrapped in a new method, _process_file_.

We need to test that files are being recompiled or not correctly. We have unit tests for the recompile logic but we should also have a system test for this. These are currently in https://github.com/bblay/fab/pull/12 .


source/fab/steps/preprocess.py
```diff
@@ -126,6 +126,7 @@ def fortran_preprocessor(preprocessor=None, source=None,
     Params as per :class:`~fab.steps.preprocess.PreProcessor`.
 
     """
+    # todo: we want to add -P ... IF it's not already there
```
MatthewHambley, line 129: This makes assumptions about what arguments the preprocessor takes. Ultimately there needs to be way of representing each tool implementation with a standardised set of data describing the correct arguments to pass for various functions.

source/fab/util.py
```diff
@@ -48,13 +50,31 @@ def log_or_dot_finish(logger):
...
-def do_checksum(fpath: Path):
+def file_checksum(fpath):
     """
-    Checksum contents of a file.
+    Return a checksum of the given file.
+
+    This function is deterministic, returning the same result across Python invocations.
+
+    We use crc32 for now because it's deterministic, unlike out-the-box hash.
+    We could seed hash with a non-random or look into hashlib, if/when we want to improve this.
```
MatthewHambley, line 60: Might consider Adler32 as it is faster than CRC32 and almost as expressive. Or it may not matter.

source/fab/util.py
```diff
@@ -127,18 +147,67 @@ class CompiledFile(object):
...
+        :param input_fpath:
             The file that was compiled.
         :param output_fpath:
-            The object file.
+            The object file that was created.
+
+        """
+        # todo: Should just be the input_fpath, not the whole analysed file
+        self.input_fpath = Path(input_fpath)
+        self.output_fpath = Path(output_fpath)
```
MatthewHambley, line 161: Is there a reason why these are not just passed as `Path` objects?

source/fab/util.py
```diff
@@ -127,18 +147,67 @@ class CompiledFile(object):
...
+
+        self.source_hash = source_hash or 0
+        self.flags_hash = flags_hash or 0
```
MatthewHambley, line 164: Check that hashing function can't produce `0` from data input.

source/fab/util.py
```diff
@@ -127,18 +147,67 @@ class CompiledFile(object):
...
+        self.module_deps_hashes = module_deps_hashes or {}
+
+    #
+    # persistence
+    #
+    @classmethod
+    def field_names(cls):
```
MatthewHambley, line 171: Names for which fields? Presumably the persistence database. It would be good to properly encapsulate the persistance at some point.

source/fab/util.py
```diff
@@ -127,18 +147,67 @@ class CompiledFile(object):
...
+            module_deps_hashes = [i.split('=') for i in d["module_deps_hashes"].split(';')]
+            module_deps_hashes = {i[0]: int(i[1]) for i in module_deps_hashes}
+        else:
+            module_deps_hashes = {}
+
+        return cls(
+            input_fpath=Path(d["input_fpath"]),
+            output_fpath=Path(d["output_fpath"]),
+            source_hash=int(d["source_hash"]),
+            flags_hash=int(d["flags_hash"]),
```
MatthewHambley, line 205: Duplication again.

system-tests-new/tiny-project/test_tiny_project.py
```diff
@@ -0,0 +1,56 @@
...
+from fab.steps.preprocess import fortran_preprocessor
+
+
+class TestTinyProject(object):
+
+    def test(self, tmp_path):
+        # We want to build in a clean workspace each time the test is run.
+        # (Pytest provides the tmp_path fixture.)
+        self.fab_workspace = tmp_path
+        print(f"fab_workspace is {self.fab_workspace}")
```
MatthewHambley, line 19: Do we want to be printing out of the tests?

system-tests-new/tiny-project/test_tiny_project.py
```diff
@@ -0,0 +1,56 @@
...
+    def clean_build(self):
+        this_folder = Path(__file__).parent
+
+        config = BuildConfig(
+            project_label='tiny project',
+            fab_workspace=self.fab_workspace,
+            steps=[
+                GrabFolder(this_folder / 'project-source', dst='src'),
+                FindSourceFiles(),
+                fortran_preprocessor(preprocessor='cpp -traditional-cpp -P'),
```
MatthewHambley, line 36: Concerned about making assumptions regarding available tools. On the other hand testing is about limiting the number of variables. Is there a way of testing which tools are available and running only applicable tests?

system-tests-new/tiny-project/test_tiny_project.py
```diff
@@ -0,0 +1,56 @@
...
+                Analyse(root_symbol='my_prog'),
+                CompileFortran(compiler='gfortran -c', common_flags=['-J', '$output']),
+                LinkExe(flags=['-lgfortran']),
+            ]
+        )
+
+        config.run()
+
+        # check it built ok
```
MatthewHambley, line 45: Yes, do this.

tests/integration_tests/test_compilation_results_file.py
```diff
@@ -0,0 +1,30 @@
...
+
+from fab.util import CompiledFile
+
+from fab.steps.compile_fortran import CompileFortran
+
+
+def test_compilation_results(tmp_path):
+
+    # write a few compilation results to file
+    config = mock.Mock(project_workspace=Path(tmp_path))
```
MatthewHambley, line 12: Not sure what `Mock` is doing here. Doesn't simple variable assignment allow for changing the project workspace path?

tests/unit_tests/steps/test_compile_fortran.py
```diff
@@ -29,49 +31,172 @@ def artefact_store(analysed_files):
...
+        assert artefact_store == {
+            OBJECT_FILES: {
+                'root1': {Path('root1.o'), Path('dep1.o')},
+                'root2': {Path('root2.o'), Path('dep2.o')},
+            }
+        }
+
+
+class Test_process_file(object):
+    pass
```
MatthewHambley, line 110: Missing test?

MatthewHambley (changes requested):
A reasonably substantial change, as expected for new functionality. A number of questions to answer and a few bits and bobs to tidy up.

source/fab/build_config.py
```diff
@@ -52,17 +52,23 @@ def __init__(self, project_label: str, source_root: Optional[Path] = None, steps
...
-        else:
-            fab_workspace_root = Path(os.path.expanduser("~/fab-workspace"))
-            logger.info(f"FAB_WORKSPACE not set. Using {fab_workspace_root}.")
-        self.project_workspace = fab_workspace_root / (project_label.replace(' ', '-'))
+        if not fab_workspace:
+            if os.getenv("FAB_WORKSPACE"):
+                fab_workspace = Path(os.getenv("FAB_WORKSPACE"))  # type: ignore
+            else:
+                fab_workspace = Path(os.path.expanduser("~/fab-workspace"))
+                logger.info(f"FAB_WORKSPACE not set. Defaulting to {fab_workspace}.")
```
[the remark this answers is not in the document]
bblay: I think it's useful as a reminder that the user might not have pointed Fab to the fastest drive, but happy to remove if preferred.

source/fab/build_config.py
```diff
@@ -216,6 +222,8 @@ def __init__(self, common_flags=None, path_flags: List[AddFlags] = None):
     #       make sure it's all properly documented and rationalised.
     def flags_for_path(self, path: Path, source_root: Path, project_workspace: Path):
         """
+        Get all the flags for a given file, in a reproducible order.
+
```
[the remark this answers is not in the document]
bblay: It's interesting to the reader of the incremental build code, which takes a hash of the returned flags and so needs them in the same order each time.

source/fab/steps/compile_fortran.py
```diff
@@ -52,12 +64,21 @@ def __init__(self, compiler: str = None, common_flags: List[str] = None, path_fl
             Human friendly name for logger output, with sensible default.
 
         """
+
+        # todo: perhaps this should add the -J (or -modules) automatically
+
         compiler = compiler or os.getenv('FC', 'gfortran -c')
```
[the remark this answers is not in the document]
bblay: Yes, this is certainly something we need to discuss. I don't have a preference. I guess `gfortran -c` could be also considered the command to compile a file. There's the `traditional-cpp` and '-P' flags too, which I think currently live both in the command and in flags, in various places. Perhaps we should be flexible and allow the user to have either, as they might be beholden to their environment's FC variable.

source/fab/steps/compile_fortran.py
```diff
@@ -70,111 +91,212 @@ def run(self, artefact_store, config):
...
+    def get_compile_next(self, compiled: Dict[Path, CompiledFile], uncompiled: Set[AnalysedFile]) -> Set[AnalysedFile]:
 
         # find what to compile next
         compile_next = set()
         not_ready: Dict[Path, List[Path]] = {}
-        for af in to_compile:
+        for af in uncompiled:
             # all deps ready?
-            unfulfilled = [dep for dep in af.file_deps if dep not in already_compiled_files and dep.suffix == '.f90']
+            unfulfilled = [dep for dep in af.file_deps if dep not in compiled and dep.suffix == '.f90']
```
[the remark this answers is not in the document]
bblay: It might be me who is missing something. Let's explore!

If a file depends on a module which doesn't exist, then it needs Fab to compile the file which produces that module. Fab detects module dependencies and turns them into file dependencies for the compilation stage, as the only thing that matters is the order in which the files are compiled.

With this incremental build PR, and _a.f90 --> b.mod_, Fab can tell if _module b_ has changed since _file a_ was last compiled.

Not sure if that was helpful!

source/fab/steps/compile_fortran.py
```diff
@@ -70,111 +91,212 @@ def run(self, artefact_store, config):
...
+    def recompile_check(self, analysed_file: AnalysedFile, flags_hash: int, last_compile: Optional[CompiledFile]):
+
+        # todo: other environmental considerations for the future:
+        #   - env vars, e.g OMPI_FC
+        #   - compiler version
+
+        recompile_reasons = []
+        # first encounter?
+        if not last_compile:
+            recompile_reasons.append(NO_PREVIOUS_RESULT)
```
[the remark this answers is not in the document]
bblay: That's true. We chose to give the user _all_ the reasons instead of just the first, as it seemed more useful to them, and the time saving seems to be negligible. Happy to reconsider.

source/fab/steps/compile_fortran.py
```diff
@@ -70,111 +91,212 @@ def run(self, artefact_store, config):
...
+            if flags_hash != last_compile.flags_hash:
+                recompile_reasons.append(FLAGS_CHANGED)
+
+            # have any of the modules on which we depend changed?
+            module_deps_hashes = {mod_dep: self._mod_hashes[mod_dep] for mod_dep in analysed_file.module_deps}
+            if module_deps_hashes != last_compile.module_deps_hashes:
+                recompile_reasons.append(MODULE_DEPENDENCIES_CHANGED)
+
+            # is the object file still there?
+            obj_file = analysed_file.fpath.with_suffix('.o')
```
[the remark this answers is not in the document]
bblay: Agreed. There's four of those already. Property seems the right way, too.

source/fab/steps/compile_fortran.py
```diff
@@ -70,111 +91,212 @@ def run(self, artefact_store, config):
...
+            if not obj_file.exists():
```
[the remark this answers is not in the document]
bblay: I'll attempt to push back on this, if I may, for two reasons: Firstly, as discussed above, it seems better to inform the user _all_ the reasons triggering a recompile, so both the object and hashes will be checked. Secondly, I've tried to arrange these checks in the order in which the user percieved things happen, to hopefully help the reader reason about them a little easier. Still happy to rearrange if preferred, of course!

source/fab/steps/compile_fortran.py
```diff
@@ -70,111 +91,212 @@ def run(self, artefact_store, config):
...
+                recompile_reasons.append(OBJECT_FILE_NOT_PRESENT)
+
+            # are the module files we define still there?
+            build_output = self._config.project_workspace / BUILD_OUTPUT
+            mod_def_files = [build_output / f'{mod}.mod' for mod in analysed_file.module_defs]
```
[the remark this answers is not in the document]
bblay: That sounds alright. In this case it won't know the _build_output_ folder is but that's probably ok.

source/fab/steps/compile_fortran.py
```diff
@@ -70,111 +91,212 @@ def run(self, artefact_store, config):
...
+            if not all([mod.exists() for mod in mod_def_files]):
+                recompile_reasons.append(MODULE_FILE_NOT_PRESENT)
+
+        return ", ".join(recompile_reasons)
```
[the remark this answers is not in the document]
bblay: I just imagined being the user, _"wait, the flags changed too?!"_ or something. Since we can tell the user everything that's changed, it would seem potentially more useful to do so. Happy to reconsider.

source/fab/steps/preprocess.py
bblay: Yes, that definitely needs addressing. Perhaps a chat sometime soon?

source/fab/util.py
bblay: I think it doesn't matter because I've seen Fab take hashes of 3-4k files in <0.1s.

source/fab/util.py
bblay: Future robustness/flexibility in case a user-defined step starts adding CompiledFile objects and doesn't pass in paths. Always happy to reconsider.

source/fab/util.py
bblay: It's only 0 when we're not using a hashing function at all, as in the case of the C compiler (or tests), so there can be no clash.

source/fab/util.py
bblay: Agreed - both for this and the analysis data. Disappointed [dbm](https://docs.python.org/3/library/dbm.html) wasn't up to the job!

source/fab/util.py
```diff
@@ -127,18 +147,67 @@ class CompiledFile(object):
...
-        # todo: A compiled file shouldn't know whether its source file was analysed or not.
-        #       It needn't be in some use cases. This is a code smell and needs revisiting.
-        self.analysed_file = analysed_file
-        self.output_fpath = output_fpath
+        return {
+            "input_fpath": str(self.input_fpath),
+            "output_fpath": str(self.output_fpath),
+            "source_hash": str(self.source_hash),
+            "flags_hash": str(self.flags_hash),
+            "module_deps_hashes": ';'.join([f'{k}={v}' for k, v in self.module_deps_hashes.items()]),
```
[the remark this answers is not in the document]
bblay: I don't understand, sorry.

system-tests-new/tiny-project/test_tiny_project.py
bblay: That print is useful when a test is failing and we want to investigate the artefacts.

(Also, it was suggested that we pull out these tests into a separate PR, which I've done here: https://github.com/bblay/fab/pull/12.)

system-tests-new/tiny-project/test_tiny_project.py
bblay: That's an interesting question which might also useful beyond testing. Chat?

tests/unit_tests/steps/test_compile_fortran.py
bblay: Yes, thank you!

system-tests-new/tiny-project/test_tiny_project.py
```diff
@@ -0,0 +1,56 @@
...
+        # We want to build in a clean workspace each time the test is run.
+        # (Pytest provides the tmp_path fixture.)
+        self.fab_workspace = tmp_path
+        print(f"fab_workspace is {self.fab_workspace}")
+
+        self.clean_build()
+
+        # self.no_change_rebuild()
+        #
+        # self.incremental_build()
```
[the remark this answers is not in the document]
bblay: (now a separate PR)

tests/integration_tests/test_compilation_results_file.py
bblay: That works too. Will do.

source/fab/steps/compile_fortran.py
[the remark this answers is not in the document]
MatthewHambley: I was likening it to the previously discussed `thing.o` situation. It seems like the analysed file should be able to return a list of module files rather than expecting the caller to do a `.withsuffix` or in this case `f'{mod}.mod}`.

On a passing note, are you aware that the Cray compiler (which we will need to support) generates `*.MOD` not `*.mod`?

source/fab/steps/compile_fortran.py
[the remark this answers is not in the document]
MatthewHambley: That seems fine with the exception that if a file is just missing then being told that it doesn't match hashes and such like isn't very helpful. It's implied by the absence of the file.

source/fab/util.py
MatthewHambley: We will need to support incremental building of C code as well. I'm just mentioning this in case that had slipped through the cracks. It doesn't have to be done in this change though.

source/fab/util.py
MatthewHambley: In the meantime can you rename this function just to make it clear what is being named.

source/fab/util.py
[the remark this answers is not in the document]
MatthewHambley: Function `field_names` lists all the field names. Then `to_str_dict` repeats all that data again. Then this function repeats it all again. It would be good to have a single source of truth for the field names.

MatthewHambley (changes requested):
Just a few minor things to close down.

source/fab/steps/compile_fortran.py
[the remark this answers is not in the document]
bblay: Will do.

`*.MOD`? No I wasn't, thank you.

source/fab/steps/compile_fortran.py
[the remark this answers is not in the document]
bblay: Fab doesn't try to compile files that don't exist.
Do you mean if a mod is missing we shouldn't tell them it's hash has changed?
We don't do that, we only check _mod dependency hashes_ and _mod definition existence_.

source/fab/util.py
bblay: Agreed. Of course when we do they won't default to 0.

source/fab/util.py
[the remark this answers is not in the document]
bblay: I see, thanks. The best solution to that is probably to have a `fields` method which returns `Dict[name, type]`. Because `to_str_dict` and `from_str_dict` would become generic, it would seem well placed in a _"Dictable"_ (or better named) base class. This is quite nice. I can do it here or in a follow-up ticket - but aren't we likely to remove all this with a move to a database sometime soon?

MatthewHambley (approved):
Lets get this on so it can be polished and more extensively tested.

PR #133, opened 2022-08-02
System tests

bblay:
Very basic system tests which run the example run configs.

We'll find a way to run this from master, in a cron.


bblay:
Closed because merged into #134.

PR #134, opened 2022-08-09
Compile mods first

bblay:
[Busby-inspired](https://www.osti.gov/biblio/1393322) two-stage Fortran compilation, with:
 - Example run configs updated with `--two-stage` and `-opt` args.
 - A system test which runs the configs with/without optimisation and two stages.

Only tested with gfortran so far. The tests are becoming increasingly gfortran-oriented. This needs addressing in the future.


bblay:
Closing for now, I think there's a mistake I need to fix before review.

PR #137, opened 2022-08-22
fix call analysis for lfric method calls

bblay:
The new call analysis code didn't work with `%` method calls.

It wasn't spotted because we only checked some of the run configs. The work we're currently doing on system tests should help avoid this class of problem in the future.


MatthewHambley (approved):
According to the developer the caller will be unnamed if it is a function call or object method. In those cases prerequisites are already taken care of by the `use` statements.

PR #140, opened 2022-09-01
pragma injector tidy

bblay:
move pragma injector and test into expected place


MatthewHambley (approved):
A simple relocation of functionality. Nothing controversial here.

PR #141, opened 2022-09-06
unit tests for compile_c

bblay:
(although they're quite integration-ey, rather than unit-ey)


MatthewHambley (approved):
This looks fine.

PR #142, opened 2022-09-07
github system tests

bblay:
Runs systems-tests-new from the github build.yml.
 - Includes a Dockerfile for running the system tests on Windows (through PyCharm).
 - One system test had a flawed assert which slipped through the net, ~~and which previously passed due to a non-idempotent function (which didn't affect functionality)~~. This highlights a fragility in the method of testing, which compares csv contents in a concise way. We hope/expect to retire CSVs in favour a third-party database soon, where this concern will go away.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,3 +48,7 @@ jobs:
       run: |
         pip install .[tests]
         python -m pytest --cov=fab tests/integration_tests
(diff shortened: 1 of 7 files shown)


MatthewHambley (approved):
This seems to do the necessary. System tests are now run (and run to completion) as part of the commit workflow.

PR #143, opened 2022-09-09
Remove MpExe class

bblay:
Remove the spurious MpExe base class, which seemed useful at the time but really wasn't. Steve Wardle rightly pointed this out before he left.


docs/source/overview.rst
```diff
@@ -71,12 +71,9 @@ Under consideration: Fab might get a config file format in the future.
...
 ---------------
 
-Steps have access to multiprocessing methods. The :class:`~fab.steps.mp_exe.MpExeStep` class captures common aspects
-of any step that passes a list of artefacts through a command-line tool.
-(The existence of this class is currently under debate.)
-The Step class includes a multiprocessing utility method called :meth:`~fab.steps.Step.run_mp` which steps can call
-from their :meth:`~fab.steps.Step.run` method to pass a collection of artefacts through a processor function
-in parallel.
+Steps have access to multiprocessing methods.
+The Step class includes a multiprocessing helper method called :meth:`~fab.steps.Step.run_mp` which steps can call
```
GiorgiaLine, line 75: You might want to reconsider the name of this helper method, because I guarentee a few people will be unhappy with it.

bblay: Oh, thanks @GiorgiaLine. It's not clear which kind of names we need to avoid here or what the objection would be in this case. Are you able to elaborate please? Does it help that _:meth:`~fab.steps.Step.run_mp`_ is rendered by Sphinx as just a clickable `run_mp`, and that a step writer would access the function as `self.run_mp`?

GiorgiaLine: Never mind, coming back to this with fresh eyes I realise I've misread it and the variable name itself isn't meth.

source/fab/steps/compile_fortran.py
```diff
@@ -70,9 +70,10 @@ def __init__(self, compiler: str = None, common_flags: List[str] = None, path_fl
         """
 
         # todo: perhaps this should add the -J (or -modules) automatically
+        super().__init__(name=name)
 
-        compiler = compiler or os.getenv('FC', 'gfortran -c')
-        super().__init__(exe=compiler, common_flags=common_flags, path_flags=path_flags, name=name)
+        self.exe = compiler or os.getenv('FC', 'gfortran -c')
```
bblay, line 75: Call this 'compiler'. Do similar in preprocessor.

GiorgiaLine (approved):
Changes look good, ready for merge.

source/fab/steps/compile_fortran.py
bblay: (addressed elsewhere)

PR #144, opened 2022-09-09
restore system tests: FortranDependencies

bblay:
Now checks the output of both exes. Also:
 - The analysis file loading has been pulled out of the `Analyse` class into a standalone function, `_load_analysis_file`, so it can be tested more easily. It makes sense to do this anyway. I think this new function and the `AnalysedFile` class might end up in their own module. Please shout of you want that explored in this PR.
 - Adds a _.gitattributes_ file to stop the sample source code being checked out with `\r\n` line feeds on Windows, which was changing the crc32 results in the tests.


source/fab/steps/analyse.py
```diff
@@ -299,31 +299,25 @@ def _get_file_checksums(self, fpaths: Iterable[Path]) -> Dict[Path, int]:
...
-                    # file no longer there?
-                    if analysed_file.fpath not in latest_file_hashes:
-                        logger.info(f"file no longer present: {analysed_file.fpath}")
-                        continue
+        Reads the analysis csv file, discarding results from files which are no longer present.
 
-                    # ok, we have previously analysed this file
-                    prev_results[analysed_file.fpath] = analysed_file
+        :param fpath:
+            Path to analysis csv file.
```
MatthewHambley, line 307: There doesn't seem to be an `fpath` argument anymore.

source/fab/steps/analyse.py
```diff
@@ -437,3 +431,22 @@ def _add_unreferenced_deps(self, symbols: Dict[str, Path],
             # add the file and it's file deps
             sub_tree = extract_sub_tree(source_tree=all_analysed_files, root=analysed_fpath)
             build_tree.update(sub_tree)
+
+
+def _load_analysis_file(fpath) -> Dict[Path, AnalysedFile]:
```
MatthewHambley, line 436: This function, on the other hand, does have an `fpath` argument. Have they got mixed up? Also, does `fpath` have a type?

source/fab/steps/analyse.py
```diff
@@ -437,3 +431,22 @@ def _add_unreferenced_deps(self, symbols: Dict[str, Path],
...
+    # Return the contents of an analysis csv.
+    results: Dict[Path, AnalysedFile] = dict()
+    try:
+        with open(fpath, "rt") as csv_file:
+            dict_reader = csv.DictReader(csv_file)
+
+            # read every row and convert into an AnalysedFile object
+            for row in dict_reader:
+                analysed_file = AnalysedFile.from_str_dict(row)
+                results[analysed_file.fpath] = AnalysedFile.from_str_dict(row)
```
MatthewHambley, line 446: Isn't an `AnalysedFile` object being created twice for the same row? Should the `analysed_file` object be used instead of creating a new object?

MatthewHambley (changes requested):
Although the expected output files have been moved to the new system test directory I don't see any evidence of them being used in the test.

I don't think there's any need to complicate this patch with migrating code to a new modules. That will make a separate patch.

source/fab/steps/analyse.py
bblay: Yes, they were one function, now split. Thanks.

source/fab/steps/analyse.py
bblay: Yes, sorry. I suspect this might have been JetBrains' magic wand, which we've been using with glee to resolve conflicts where possible. I recently noticed, for the first time ever, that it didn't behave as expected so we'll keep a closer eye on this from now on, and maybe do it old skool instead.

MatthewHambley (approved):
This change seems ready to go.

PR #145, opened 2022-09-15
Fortran prebuild concept

bblay:
A simple Fortran pre-build solution, somewhere between a code-sketch and a proposal.
 - The user can copy other users' pre-built object files onto their machine without worrying about overwriting something they need.
 - The location of the object files is _above_ the project workspace, in the _Fab_ workspace, and is also flat.
 - A further improvement, again seemingly simple, would be to allow the user to merely point to other users' workspaces, from which Fab will only copy the files it needs.
 - There may be file proliferation over time, which is trivial to solve and ignored here.

### DIFF
--- a/run_configs/jules/build_jules.py
+++ b/run_configs/jules/build_jules.py
@@ -8,13 +8,15 @@ import logging
 import os
 from argparse import ArgumentParser
 
+from fab.constants import PREBUILD
+
 from fab.steps.archive_objects import ArchiveObjects
 
 from fab.build_config import BuildConfig
 from fab.steps.analyse import Analyse
 from fab.steps.compile_c import CompileC
 from fab.steps.compile_fortran import CompileFortran
-from fab.steps.grab import GrabFcm
+from fab.steps.grab import GrabFcm, GrabPreBuild
 from fab.steps.link_exe import LinkExe
 from fab.steps.preprocess import c_preprocessor, fortran_preprocessor
 from fab.steps.root_inc_files import RootIncFiles
@@ -44,6 +46,9 @@ def jules_config(revision=None, two_stage=False, opt='Og'):
 
         GrabFcm(src='fcm:jules.xm_tr/src', revision=revision, dst='src'),
         GrabFcm(src='fcm:jules.xm_tr/utils', revision=revision, dst='utils'),
+        
+        # Copy another pre-build folder into our own.
+        GrabPreBuild(path=f'/home/ho6/dbrown/fab_workspace/{config.project_label}/{PREBUILD}', allow_fail=True),
 
         FindSourceFiles(path_filters=[
             Exclude('src/control/um/'),
--- a/run_configs/jules/build_jules.py
+++ b/run_configs/jules/build_jules.py
@@ -48,7 +48,8 @@ def jules_config(revision=None, two_stage=False, opt='Og'):
         GrabFcm(src='fcm:jules.xm_tr/utils', revision=revision, dst='utils'),
         
         # Copy another pre-build folder into our own.
-        GrabPreBuild(path=f'/home/ho6/dbrown/fab_workspace/{config.project_label}/{PREBUILD}', allow_fail=True),
+        # GrabPreBuild(path=f'/home/ho6/dbrown/fab_workspace/{config.project_label}/{PREBUILD}', allow_fail=True),
+        GrabPreBuild(path=f'/home/h02/bblay/temp_prebuild', allow_fail=True),
 
         FindSourceFiles(path_filters=[
             Exclude('src/control/um/'),
--- a/source/fab/build_config.py
+++ b/source/fab/build_config.py
@@ -18,7 +18,7 @@ from pathlib import Path
 from string import Template
 from typing import List, Optional, Dict, Any
 
-from fab.constants import BUILD_OUTPUT, SOURCE_ROOT
+from fab.constants import BUILD_OUTPUT, SOURCE_ROOT, ARTEFACT_REPOSITORY_FOLDER
 from fab.metrics import send_metric, init_metrics, stop_metrics, metrics_summary
 from fab.steps import Step
 from fab.util import TimerLogger, by_type
@@ -69,6 +69,7 @@ class BuildConfig(object):
                 logger.info(f"FAB_WORKSPACE not set, defaulting to {fab_workspace}")
         logger.info(f"\nfab workspace is {fab_workspace}")
 
+        self.fab_workspace = fab_workspace
         self.project_workspace = fab_workspace / (project_label.replace(' ', '-'))
         self.metrics_folder = self.project_workspace / 'metrics' / self.project_label.replace(' ', '_', -1)
 
@@ -95,6 +96,10 @@ class BuildConfig(object):
     @property
     def build_output(self):
         return self.project_workspace / BUILD_OUTPUT
+
+    @property
+    def artefact_repository_folder(self):
+        return self.fab_workspace / ARTEFACT_REPOSITORY_FOLDER
 
     def run(self):
         """
--- a/source/fab/build_config.py
+++ b/source/fab/build_config.py
@@ -18,7 +18,7 @@ from pathlib import Path
 from string import Template
 from typing import List, Optional, Dict, Any
 
-from fab.constants import BUILD_OUTPUT, SOURCE_ROOT, ARTEFACT_REPOSITORY_FOLDER
+from fab.constants import BUILD_OUTPUT, SOURCE_ROOT, PREBUILD
 from fab.metrics import send_metric, init_metrics, stop_metrics, metrics_summary
 from fab.steps import Step
 from fab.util import TimerLogger, by_type
@@ -34,7 +34,7 @@ class BuildConfig(object):
 
     def __init__(self, project_label: str, source_root: Optional[Path] = None, steps: Optional[List[Step]] = None,
                  multiprocessing=True, n_procs: int = None, reuse_artefacts=False,
-                 fab_workspace: Optional[Path] = None, verbose=False):
+                 fab_workspace: Optional[Path] = None, verbose=False, prebuild_folder=None):
         """
         :param str project_label:
             Name of the build project. The project workspace folder is created from this name, with spaces replaced
@@ -56,9 +56,11 @@ class BuildConfig(object):
         :param fab_workspace:
             Overrides the FAB_WORKSPACE environment variable.
             If not set, and FAB_WORKSPACE is not set, the fab workspace defaults to *~/fab-workspace*.
+        :param prebuild_folder:
+            Optionally specify the pre-build folder. Defaults to <project workspace>/build_output/prebuilds.
 
         """
-        self.project_label: str = project_label
+        self.project_label: str = project_label.replace(' ', '_')
 
         # workspace folder
         if not fab_workspace:
@@ -69,12 +71,12 @@ class BuildConfig(object):
                 logger.info(f"FAB_WORKSPACE not set, defaulting to {fab_workspace}")
         logger.info(f"\nfab workspace is {fab_workspace}")
 
-        self.fab_workspace = fab_workspace
-        self.project_workspace = fab_workspace / (project_label.replace(' ', '-'))
-        self.metrics_folder = self.project_workspace / 'metrics' / self.project_label.replace(' ', '_', -1)
+        self.project_workspace = fab_workspace / self.project_label
+        self.metrics_folder = self.project_workspace / 'metrics' / self.project_label
 
         # source config
         self.source_root: Path = source_root or self.project_workspace / SOURCE_ROOT
+        self.prebuild_folder = prebuild_folder or self.build_output / PREBUILD
 
         # build steps
         self.steps: List[Step] = steps or []
@@ -96,10 +98,6 @@ class BuildConfig(object):
     @property
     def build_output(self):
         return self.project_workspace / BUILD_OUTPUT
-
-    @property
-    def artefact_repository_folder(self):
-        return self.fab_workspace / ARTEFACT_REPOSITORY_FOLDER
 
     def run(self):
         """
--- a/source/fab/steps/compile_fortran.py
+++ b/source/fab/steps/compile_fortran.py
@@ -10,6 +10,7 @@ Fortran file compilation.
 import csv
 import logging
 import os
+import zlib
 from collections import defaultdict
 from pathlib import Path
 from typing import List, Set, Dict, Optional
@@ -79,7 +80,7 @@ class CompileFortran(MpExeStep):
 
         # runtime
         self._stage = None
-        self._last_compiles: Dict[Path, CompiledFile] = {}
+        # self._last_compiles: Dict[Path, CompiledFile] = {}
         self._mod_hashes: Dict[str, int] = {}
 
     def run(self, artefact_store, config):
@@ -99,7 +100,7 @@ class CompileFortran(MpExeStep):
         super().run(artefact_store, config)
 
         # read csv of last compile states
-        self._last_compiles = self.read_compile_result(config)
+        # self._last_compiles = self.read_compile_result(config)
 
         # get all the source to compile, for all build trees, into one big lump
         build_lists: Dict[str, List] = self.source_getter(artefact_store)
@@ -129,7 +130,7 @@ class CompileFortran(MpExeStep):
             compiled_this_pass = list(by_type(results_this_pass, CompiledFile))
             logger.info(f"stage 2 compiled {len(compiled_this_pass)} files")
 
-        self.write_compile_result(compiled, config)
+        # self.write_compile_result(compiled, config)
 
         self.store_artefacts(compiled, build_lists, artefact_store)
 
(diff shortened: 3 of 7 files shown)


source/fab/steps/compile_fortran.py
```diff
@@ -193,142 +184,89 @@ def store_artefacts(self, compiled_files: Dict[Path, CompiledFile], build_trees:
...
-        * hash of the compile flags
-        * hash of the module files on which we depend
+        Returns a compilation result, regardless of whether it was compiled or prebuilt.
 
         """
-        # flags for this file
+        # todo: include compiler version in hashes
+
+        # get a combo hash of things which matter to the mod files we define
+        if analysed_file.file_hash is None:
```
bblay, line 197: These asserts were added during debugging. I think this error handling should instead become a `try` around the `sum`.

source/fab/steps/compile_fortran.py
```diff
@@ -193,142 +184,89 @@ def store_artefacts(self, compiled_files: Dict[Path, CompiledFile], build_trees:
...
     def process_file(self, analysed_file: AnalysedFile):
         """
         Prepare to compile a fortran file, and compile it if anything has changed since it was last compiled.
 
-        Returns a compilation result, including various hashes from the time of compilation:
-
-        * hash of the source file
-        * hash of the compile flags
-        * hash of the module files on which we depend
+        Returns a compilation result, regardless of whether it was compiled or prebuilt.
```
bblay, line 191: The docstring should say something about how it copies into / out of the prebuild folder.

source/fab/steps/compile_fortran.py
```diff
@@ -193,142 +184,89 @@ def store_artefacts(self, compiled_files: Dict[Path, CompiledFile], build_trees:
...
 
             # files
-            command.append(str(analysed_file.fpath))
+            command.append(analysed_file.fpath.name)
             command.extend(['-o', str(output_fpath)])
 
             log_or_dot(logger, 'CompileFortran running command: ' + ' '.join(command))
-            run_command(command)
+
+            run_command(command, cwd=analysed_file.fpath.parent)
```
bblay, line 265: The docstring should mention it does this.

source/fab/steps/compile_fortran.py
```diff
@@ -69,17 +60,22 @@ def __init__(self, compiler: str = None, common_flags: List[str] = None, path_fl
...
-        # todo: perhaps this should add the -J (or -modules) automatically
-
+        # todo: Fab should be compiler-aware
         compiler = compiler or os.getenv('FC', 'gfortran -c')
-        super().__init__(exe=compiler, common_flags=common_flags, path_flags=path_flags, name=name)
+        common_flags = common_flags or []
+        env_flags = os.getenv('FFLAGS', '').split()
+        super().__init__(
+            exe=compiler,
+            common_flags=remove_minus_J(common_flags + env_flags, verbose=True),
```
bblay, line 69: The docstring should say it does this, and why.

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
```diff
@@ -0,0 +1,84 @@
...
+            my_config.run()
+
+        # make sure we didn't call the compiler
+        mock_compile_file.assert_not_called()
+
+        # make sure the exe was built
+        exe = my_config.project_workspace / 'my_prog.exe'
+        assert exe.exists()
+
+        print(tmp_path)
```
bblay, line 83: remove prints

system-tests-new/run-configs/run-configs.sh
```diff
@@ -12,21 +12,21 @@ echo $(pwd)
 echo
 echo "building gcom variants"
 echo
-cd gcom
-  ./grab_gcom.py
-  ./build_gcom_ar.py
-  ./build_gcom_so.py
-cd ..
+#cd gcom
```
bblay, line 15: restore these, commented our for convenience and forgotten about

tests/unit_tests/steps/test_compile_fortran.py
```diff
@@ -109,129 +109,164 @@ def test_vanilla(self, compiler):
...
+        with mock.patch('pathlib.Path.exists', return_value=True):  # mod def files and obj file all exist
+            with mock.patch('fab.steps.compile_fortran.CompileFortran.compile_file') as mock_compile_file:
+                with mock.patch('shutil.copy2') as mock_copy:
+                    res = compiler.process_file(analysed_file)
+
+        assert res == CompiledFile(input_fpath=analysed_file.fpath, output_fpath=expect_object_fpath)
+        mock_compile_file.assert_not_called()
+        self.ensure_mods_restored(mock_copy)
+
+    def test_file_hash(self):
```
bblay, line 182: Some of these tests might be better if they used real files instead of mocking the existence of the expected filenames.

bblay (commented):
I gave it a once-over as I was a bit poorly when I wrote this.

source/fab/steps/grab.py
```diff
@@ -157,3 +155,40 @@ def run(self, artefact_store: Dict, config):
 
             r = remote.RemoteClient(self.src)
             r.export(str(config.source_root / self.dst_label), revision=self.revision, force=True)
+
+
+class GrabPreBuild(Step):
```
MatthewHambley, line 160: We may want to think about alternate naming. So far "grab" has always been used to build a source tree for compilation. Here it's being used to manage internal state.

source/fab/util.py
```diff
@@ -62,6 +62,31 @@ def file_checksum(fpath):
         return HashedFile(fpath, zlib.crc32(infile.read()))
 
 
+def remove_minus_J(flags, verbose=False):
+    """
+    Remove the -J <folder> from the given flags.
+
+    """
+    # todo: what if there's two -Js ?
+    # todo: Fab should be compiler aware, with the possibly of different flags to ignore per compiler
```
MatthewHambley, line 71: As mentioned elsewhere, I think that ideally we wouldn't have to pick apart already constructed argument lists.

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
```diff
@@ -0,0 +1,84 @@
...
+
+        config2 = self.build_config(fab_workspace=tmp_path / 'workspace_two')
+        config2.run()
+
+        files2 = set(file_walk(config2.build_output))
+        hashes2 = {f.relative_to(config2.build_output): zlib.crc32(open(f, 'rb').read()) for f in files2}
+
+        del hashes1[Path('__analysis.csv')]
+        del hashes2[Path('__analysis.csv')]
+        assert hashes1 == hashes2
```
MatthewHambley, line 60: I may be misunderstanding what is being tested but this seems to be testing that the output files have the same contents. Don't you want to test that their filenames contain the same hashes?

source/fab/build_config.py
```diff
@@ -34,7 +34,7 @@ class BuildConfig(object):
 
     def __init__(self, project_label: str, source_root: Optional[Path] = None, steps: Optional[List[Step]] = None,
                  multiprocessing=True, n_procs: int = None, reuse_artefacts=False,
-                 fab_workspace: Optional[Path] = None, verbose=False):
+                 fab_workspace: Optional[Path] = None, verbose=False, prebuild_folder=None):
```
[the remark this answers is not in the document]
bblay: [Agreed](https://github.com/metomi/fab/projects/8#card-85863895).

source/fab/build_config.py
```diff
@@ -95,7 +98,7 @@ def __init__(self, project_label: str, source_root: Optional[Path] = None, steps
             logging.getLogger('fab').setLevel(logging.DEBUG)
 
         # runtime
-        self._artefact_store: Dict[str, Any] = dict()
+        self._artefact_store: Dict[str, Any] = None  # type: ignore
```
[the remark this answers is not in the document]
bblay: We knew you were going to say that! :D
I think it's unpythonic, and that mypy shouldn't have an opinion here, and that it's ugly whichever way we satisfy mypy here, but we're happy to oblige as suggested.

source/fab/steps/compile_fortran.py
```diff
@@ -69,17 +60,22 @@ def __init__(self, compiler: str = None, common_flags: List[str] = None, path_fl
 
         """
 
-        # todo: perhaps this should add the -J (or -modules) automatically
-
+        # todo: Fab should be compiler-aware
         compiler = compiler or os.getenv('FC', 'gfortran -c')
```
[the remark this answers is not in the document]
bblay: Sounds reasonable. Zero config would presumably detect the compiler elsewhere, and pick a pre-defined instance of this class.

source/fab/steps/compile_fortran.py
```diff
@@ -193,142 +184,89 @@ def store_artefacts(self, compiled_files: Dict[Path, CompiledFile], build_trees:
...
-        # flags for this file
+        # todo: include compiler version in hashes
+
+        # get a combo hash of things which matter to the mod files we define
+        if analysed_file.file_hash is None:
+            raise ValueError(f"unexpected None file hash: {analysed_file}")
+        assert analysed_file.file_hash is not None
+        assert self.exe is not None
+
+        mod_combo_hash = sum([analysed_file.file_hash, zlib.crc32(self.exe.encode())])
```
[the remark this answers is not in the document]
bblay: Yes, it's the latter. There's an open PR which removes the pointless `MPExe` base class, which is where this name comes from. It's subclasses should replace that with meaningful names.

source/fab/steps/compile_fortran.py
```diff
@@ -193,142 +184,89 @@ def store_artefacts(self, compiled_files: Dict[Path, CompiledFile], build_trees:
...
+        mod_file_prebuilds = {self._config.prebuild_folder / f'{mod_def}.{mod_combo_hash:x}.mod'
+                              for mod_def in analysed_file.module_defs}
+
+        # get a combo hash of things which matter to the object file we define
         flags = self.flags.flags_for_path(path=analysed_file.fpath, config=self._config)
-        flags_hash = string_checksum(str(flags))
+        mod_deps_hashes = {mod_dep: self._mod_hashes.get(mod_dep, 0) for mod_dep in analysed_file.module_deps}
+        obj_combo_hash = sum([analysed_file.file_hash, flags_checksum(flags),
```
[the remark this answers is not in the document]
bblay: We'd added some exception handling so pulled that out too.

source/fab/steps/grab.py
bblay: Always up for those chats.

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
bblay: The main point of that test is to ensure that two identical _my_mod.f90_ produce mods with identical _file contents_, despite being in different folders. It's testing the fix for removing folders from mod contents. Of course it follows then that the two identical _my_prog.f90_ which use those mods must become artefacts with identical hash filenames, as they depend on the same mod content.

Did that make sense? I think the code is simple but the use cases can be complicated!

tests/unit_tests/steps/test_compile_fortran.py
bblay: We've made [a ticket](https://github.com/metomi/fab/projects/8#card-85865236) to think about this.

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
bblay: (also, it's testing both content and filenames are identical, for the reasons "described")

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def build_config(self, fab_workspace, prebuild_folder=None):
+        logging.getLogger('fab').setLevel(logging.WARNING)
+
+        build_config = BuildConfig(
+            project_label='test_prebuild',
+            fab_workspace=fab_workspace,
+            steps=[
+                GrabFolder(Path(__file__).parent / 'project-source', dst='src'),
+                *([GrabPreBuild(prebuild_folder)] if prebuild_folder else []),
```
matthewrmshin, line 27: A bit clearer this way?

```suggestion
                [GrabPreBuild(f) for f in prebuild_folder],
```

source/fab/steps/grab.py
```diff
@@ -157,3 +155,40 @@ def run(self, artefact_store: Dict, config):
...
+        self.objects = objects
+        self.allow_fail = allow_fail
+
+    def run(self, artefact_store: Dict, config):
+        dst = config.prebuild_folder
+        try:
+            res = call_rsync(src=self.src, dst=dst)
+
+            # log the number of files transferred
+            to_print = [line for line in res.decode().split('\n') if 'Number of' in line]
```
matthewrmshin, line 177:
```suggestion
            to_print = [line for line in res.decode().splitlines() if 'Number of' in line]
```

source/fab/steps/grab.py
```diff
@@ -157,3 +155,40 @@ def run(self, artefact_store: Dict, config):
...
+            logger.warning(msg)
+
+
+def call_rsync(src: Union[str, Path], dst: Union[str, Path]):
+    # we want the source folder to end with a / for rsync because we don't want it to create a sub folder
+    src = os.path.expanduser(str(src))
+    if not src.endswith('/'):
+        src += '/'
+
+    command = ['rsync', '-times', '--stats', '-ru', src, str(dst)]
```
matthewrmshin, line 193: `--times`?

```suggestion
    command = ['rsync', '--times', '--stats', '-ru', src, str(dst)]
```

source/fab/util.py
```diff
@@ -250,10 +222,12 @@ def run_command(command, env=None):
 
     """
     logger.debug(f'run_command: {command}')
-    res = subprocess.run(command, capture_output=True, env=env)
+    res = subprocess.run(command, capture_output=True, env=env, cwd=cwd)
     if res.returncode != 0:
         raise RuntimeError(f"Command failed:\n{command}\n{res.stderr.decode()}")
 
+    return res.stdout
```
matthewrmshin, line 229: This may be OK depending on what you run, but users won't be able to see the stdout on error, which may contain clues of why it fails.

source/fab/util.py
```diff
@@ -62,6 +62,31 @@ def file_checksum(fpath):
...
+    """
+    # todo: what if there's two -Js ?
+    # todo: Fab should be compiler aware, with the possibly of different flags to ignore per compiler
+    try:
+        found = flags.index('-J')
+        if verbose:
+            logger.info(f"removing -J <folder> from {flags}")
+        flags = flags[:found] + flags[found+2:]
+    except ValueError:
+        pass
```
matthewrmshin, line 78: Handle multiple `-J ITEM`?

```suggestion
    in_minus_J = False
    ret = []
    for flag in flags:
        if in_minus_J:
            in_minus_J = False
            continue
        if flag == '-J':
             in_minus_J = True
        else:
             ret.append(flag)
```

matthewrmshin (commented):
Added some comments on style and logic related issues. Let me know if the comments don't make sense. I'll aim to run/test this branch in my environment.

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
bblay: Hi, thanks but `prebuild_folder` is a path or `None`.
I'll make sure that's more clear in the docstring and type hints.

This is what we're trying to achieve with this line; it's inserting an element (or not) into a list of steps.
```
>>> [1, *[2], 3]
[1, 2, 3]
>>> [1, *[], 3]
[1, 3]
```

source/fab/steps/grab.py
bblay: Its `-t` or `--times`.
https://linux.die.net/man/1/rsync

source/fab/util.py
bblay: Good point. I'll add it to the exception.

source/fab/steps/grab.py
bblay: Wait it _wasn't_ already `--`?! Ok thanks!
That makes me think it's no longer needed for the test,
but still feels like the right thing to do, so leaving in.

system-tests-new/fortran-prebuild/test_fortran_prebuild.py
```diff
@@ -0,0 +1,81 @@
...
+                FindSourceFiles(),
+                fortran_preprocessor(preprocessor='cpp -traditional-cpp -P'),
+                Analyse(root_symbol='my_prog'),
+                CompileFortran(compiler='gfortran -c', common_flags=['-J', '$output']),
```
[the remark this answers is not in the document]
bblay: Good question. @MatthewHambley has also mentioned this. I think we're going to have a chat about this sometime soon. Would be good to have you there.

I think Fab might become compiler-aware soon, so it would know how to use gfortran to compile a file. If that happens, all strings like this will probably be replaced.

FWIW, the original thinking was Q: _"what's the command to just compile a file"_ A: _"gfortran -c"_.

source/fab/util.py
bblay: :+1:

source/fab/steps/compile_fortran.py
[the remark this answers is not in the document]
bblay: Renamed in #149.

run_configs/jules/build_jules.py
```diff
@@ -45,6 +48,9 @@ def jules_config(revision=None, two_stage=False, opt='Og'):
         GrabFcm(src='fcm:jules.xm_tr/src', revision=revision, dst='src'),
         GrabFcm(src='fcm:jules.xm_tr/utils', revision=revision, dst='utils'),
 
+        # Copy another pre-build folder into our own.
+        GrabPreBuild(path='/home/h02/bblay/temp_prebuild', allow_fail=True),
```
matthewrmshin, line 52: I assume this hard coded path will be removed at some point?

matthewrmshin (commented):
The idea of this change is reasonable and appears to work OK for the test cases.

I assume that the hard coded path in the building configuration is a temporary thing?

matthewrmshin (approved):
Understood that the hard coded path is for illustration purpose only.

MatthewHambley (approved):
As noted by the developers there are still plenty of rough edges to be sanded off but this seems like a solid foundation.

PR #146, opened 2022-09-28
Beta docs

bblay:
Mainly a new how-to guide for writing a build script, plus a few changes to make it read well.

See my fork build at https://bblay.github.io/fab/

PR #148, opened 2022-10-04
Beta docs pt2

bblay:
More docs, this time filling in the fetaures and advanced config pages, plus tweaks to the other pages.

A build of these docs can currently be seen [here](https://bblay.github.io/fab/howto_write_config.html).

Todo:
 - [x] merge #146
 - [x] merge #145, as we refer to docstrings from there


docs/source/features.rst
```diff
@@ -3,15 +3,44 @@ Features
...
+either by copying the prebuild folder or using the :class:`~fab.steps.grab.GrabPreBuild` step.
+
+
+Metrics
+=======
+Fab will record the time taken by each step and plot them on a pie chart.
+
+.. image:: img/pie.png
+    :width: 66%
+    :alt: pie chart
```
MatthewHambley, line 29: The example plot shows the utility of the feature. Is there way to get separate metrics for both passes of two pass compilation?

docs/source/features.rst
```diff
@@ -20,6 +49,25 @@ Limitations
 Dependency detection
 --------------------
 Whilst fab can automatically determine dependencies from module use statements,
-and from standalone call statements, it doesn't currently detect a dependency from a call statement
-on a single-line if statement. We can pass the analyser any dependencies which Fab can't detect,
-and they'll make their way through to the link stage.
\ No newline at end of file
+and from standalone call statements, it doesn't currently detect a dependency from a call statement on a
+single-line if statement: `IF (x .GT. 0) CALL foo()`.
```
MatthewHambley, line 53: I presume there's a ticket for fixing this. If it doesn't already have one add a reminder to change this documentation when it is resolved.

docs/source/features.rst
```diff
@@ -20,6 +49,25 @@ Limitations
...
+
+We can pass the analyser any symbol dependencies which Fab can't detect.
+The files which contain them, *and their dependencies*, will make their way through to the compile and link stages.
+This is done using the `unreferenced_deps` argument to the :class:`~fab.steps.analyse.Analyse` step.
```
MatthewHambley, line 57: Would a code snippet showing how to do this be too big?

docs/source/features.rst
```diff
@@ -20,6 +49,25 @@ Limitations
...
+   by the time we reach the analysis stage.
+ - Fortran mod files are created in a flat folder, so Fortran module names must be unique
+   by the time we reach the compile stage.
+ - C and Fortran object files are both compiled into `.o` files so there can be no duplicates,
+   such as `util.c` and `util.f90`
+   by the time we reach the compile stage.
+
+There may be duplicates earlier in the build process. For example, there may two versions of a module,
+each wrapped in a `#ifdef` so that one of them is empty after preprocessing (empty files are ignored during analysis).
+Another example is to use file filtering in the :class:`~fab.steps.find_source_files.FindSourceFiles` step.
```
MatthewHambley, line 73: Is this an example or a approach?

docs/source/howto_write_config.rst
```diff
@@ -88,14 +89,14 @@ Next we want to preprocess our source code.
...
-It was added to the :term:`Artefact Store` by the previous step.
+Thanks to Fab's sensible defaults, the Fortran preprocessor knows where to find the Fortran source code.
+It was added to the :term:`Artefact Store` by the preceding step.
 
 .. note::
     *Artefact Store*
 
-    Steps generally create and find artefacts in this in-memory dict, arranged into named collections.
-    In this case, the Fortran preprocessor automatically looks for source code in a collection named "all_source",
+    Steps generally create and find artefacts in this dict, arranged into named collections.
```
MatthewHambley, line 98: In documentation which is not verbatim code I think whole words should be preferred. So "dictionary" in this case.

docs/source/howto_write_config.rst
```diff
@@ -162,7 +162,7 @@ This argument is omitted when building a shared or static library.
 Compile and Link
 ================
 The :class:`~fab.steps.compile_fortran.CompileFortran` step creates mod and object files
```
MatthewHambley, line 164: Not in this change but "module" files.

docs/source/install.rst
```diff
@@ -85,3 +80,9 @@ will allow you to change the code without reinstalling Fab every time::
 Please be aware of some considerations when
 `using pip and conda <https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#using-pip-in-an-environment>`_
 together.
+
+Containers
+----------
+You can create a development environment for running Fab using the docker file in Fab's github repo.
+For example, PyCharm can use the interpreter inside the container and will automatically volume mount and PYTHONPATH
+Fab.
```
MatthewHambley, line 88: This is good but for production release and further beta releases we will need to be able to install the tool without using Conda. e.g. into a python venv.

What is the state of this?

MatthewHambley (changes requested):
Looks fine, just a few minor tweaks and clarifications.

docs/source/advanced_config.rst
```diff
@@ -6,17 +6,64 @@ Advanced Config
...
-Metrics
-=======
-coming soon
+Fab includes some examples of a custom step. A simple example was created for building JULES.
+The :class:`~fab.steps.root_inc_files.RootIncFiles` step copies all `.inc` files in the source tree
+into the root of the source tree, to make subsequent preprocessing flags easier to configure.
+
+That was a simple example that didn't need to interact with the :term:`Artefact Store`.
+If you need to insert a custom step between two Fab steps, you may need your custom step to create
+an :term:`Artefact Collection`, and to configure the subsequent step to use your new collection.
```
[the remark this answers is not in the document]
bblay: Well the custom step _could_ overwrite the artefact collection it reads with the artefacts it creates, to avoid having to tell the subsequent step where to find them, but that seems less desirable. I think the act of inserting a step, and therefore a new artefact collection, does fundamentally mean you're changing the flow of data between steps, i.e what the next step ingests. Happy to chat further, of course.

docs/source/advanced_config.rst
```diff
@@ -6,17 +6,64 @@ Advanced Config
...
+Many Fab steps have an optional `source` argument which overrides the default :term:`Artefacts Getter`.
```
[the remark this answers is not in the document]
bblay: :+1: How's this instead?

> That was a simple example that didn't need to interact with the `Artefact Store`.
> Sometimes, inserting a custom step means inserting a new `Artefact Collection` into the flow of data between
> steps. We can tell a subsequent step to read our new artefacts, instead of using it's default `Artefacts Getter`.
> We do this using the `source` argument, which most Fab steps accept.
> 

I wonder if the general terminology could do with a once-over, too.

docs/source/features.rst
```diff
@@ -3,15 +3,44 @@ Features
...
+Fab analyses your source code to determine the Fortran compile order,
+which means you don't have to manually specify and maintain the compile order,
+which can become problematic in a large project.
+
 
 Incremental Build
 =================
-coming soon
+Fab "watermarks" each artefact with a checksum of its inputs.
+Subsequent builds avoid reprocessing by searching for watermarks in the prebuild folder.
```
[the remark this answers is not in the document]
bblay: Every build project puts its shareable, prebuilt artefacts into a folder called `_prebuild`, in the `build_output` folder. That's what other people can copy, and that's where we copy prebuilds from other people. It's just our incremental build folder, so perhaps it should be renamed to that?

It's not a two-stage process because everything which can be reused is in that folder. The parent `build_output` folder also contains some non-shareable things but we don't search for anything in two places because everything has one place to live. Happy to be steered on this, of course.

docs/source/features.rst
bblay: The histogram and busby plots are produced for both stages. These are "sub-step" metrics, the top-level pie chart only currently shows the total for each whole step. We could always refactor the metrics to do it though.

docs/source/features.rst
bblay: There wasn't one, [made it](https://github.com/metomi/fab/projects/8#card-86192673).

docs/source/features.rst
bblay: No that's a good idea.

docs/source/features.rst
bblay: approach, thanks

docs/source/install.rst
bblay: We haven't had to use a venv for Fab yet but this is certainly something useful to do soon. https://github.com/metomi/fab/projects/8#card-86192865

PR #149, opened 2022-10-06
rename self.exe attribute

bblay:
Actioning [this review comment](https://github.com/metomi/fab/pull/145#discussion_r979864912), self.exe is renamed for better readability.


MatthewHambley (approved):
A simple change which improves readability.

source/fab/steps/compile_fortran.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, compiler: str = None, common_flags: List[str] = None, path_fl
         compiler = compiler or os.getenv('FC', 'gfortran -c')
         common_flags = common_flags or []
         env_flags = os.getenv('FFLAGS', '').split()
-        self.exe: str = compiler or os.getenv('FC', 'gfortran -c')  # type: ignore
+        self.compiler: str = compiler or os.getenv('FC', 'gfortran -c')  # type: ignore
```
MatthewHambley, line 68: We seem to have two instances of `compiler or os.getenv('FC', 'gfortran -c')`, here and line 65. Is that intended?

MatthewHambley (changes requested):
Noticed an oddity

source/fab/steps/compile_fortran.py
bblay: It's my use of the magic wand merge button again. I _did_ say I would be more vigilant with it's use...sorry!

MatthewHambley (approved):
Doubled up compiler determination fixed.

MatthewHambley (approved):
As noted by the developers, many of the changes originally in this patch have been swept up in others that there isn't much left.

PR #153, opened 2022-10-10
184 prebuild housekeeping/pruning

### LINKED ISSUE #184: housekeeping of incremental build folder
Discuss housekeeping of prebuilds folder, which has the potential to grow over time.

Some options follow. Perhaps we should implement & parameterise more than one:
 - remove anything X days older than the last access time
 - remove all but the last N versions of a file
 - optionally remove anything not used when a build completes

Update: this is in progress.

bblay:
Remove oldest artefacts from the incremental/prebuild folder based on a number of options:
 - age behind the youngest artefact
 - n oldest versions (hashes) per artefact
 - everything that wasn't part of the current build

Some review pointers:
 - The Fortran and C analysers now return the prebuild filename alongside the analysis result.
 - `CompileFortran.process_file()` now returns a list of prebuild filenames (obj + mods) alongside the compilation result.
 - Includes a little tidy-as-we-go.

This PR needs a plan review: Should we bother checking if the file system is capable of returning access time? Might be overkill. More code to carry which, perhaps, YAGNI.

todo:
 - [x] coordinate with #178 to register incremental c artefacts as "current"

Resolves #184


bblay:
Closing temporarily to investigate a problem.

source/fab/build_config.py
```diff
@@ -121,35 +125,47 @@ def run(self):
...
+
+        self._init_logging()
+        init_metrics(metrics_folder=self.metrics_folder)
+
+        # note: initialising here gives a new set of artefacts each run
+        self.init_artefact_store()
+
+        # if the user hasn't specified any cleanup of the incremental/prebuild folder,
+        # then we add a default, hard cleanup leaving only cutting-edge artefacts.
+        if not by_type(self.steps, CleanupPrebuilds):
```
bblay, line 165: Perhaps something "softer" would be more appropriate, like 1 week old files?

source/fab/steps/cleanup_prebuilds.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class CleanupPrebuilds(Step):
+    """
+    A step to delete old files from the local incremental/prebuild folder.
+
+    Assumes prebuild filenames follow the pattern: `<stem>.<hash>.<suffix>`.
+
+    """
+    def __init__(self, older_than: Optional[timedelta] = None, n_versions: int = 0, all_unused: Optional[bool] = None):
```
MatthewHambley, line 32: Can we come up with a better API? At the moment it's possible to specify none, one or more options and we don't have logic to handle all those cases.

Does Python support multiple method definition polymorphism? i.e. `def __init__(self)`, `def __init__(self, older_than: timedelta)`, `def __init__(self, n_versions: int)`, etc.

Or is the goal to support multiple modes of operation, e.g. cleanup anything older than 5 days plus anything which hasn't been used?

source/fab/steps/cleanup_prebuilds.py
```diff
@@ -0,0 +1,179 @@
...
+        if f not in current_files:
+            logger.info(f"unused {f}")
+            os.remove(f)
+            num_removed += 1
+
+    return num_removed
+
+
+# this whole function might be overkill...how likely is this, and do we care?
+def check_fs_access_time(folder=None) -> bool:
```
MatthewHambley, line 141: I think this check is a good safety net but as you observe there's quite a lot of it. It will almost certainly never be called so the real question is, how to test it? If it can't be effectively tested then we are carrying technical debt around, if it is well tested then its size is not so much of an issue.

system-tests-new/incremental/test_incremental.py
```diff
@@ -224,3 +228,158 @@ def assert_one_artefact(self, pb_keys, prebuild_groups, prebuild_folder,
             pb_fpath = next(iter(pb_group))
             assert clean_timestamps[prebuild_folder / pb_fpath] == rebuild_timestamps[prebuild_folder / pb_fpath]
             assert clean_hashes[prebuild_folder / pb_fpath] == rebuild_hashes[prebuild_folder / pb_fpath]
+
+
+class TestCleanupPrebuilds(object):
+    # Test cleanup of the incremental build artefacts
+
+    def create_artefacts(self, folder):
```
MatthewHambley, line 236: You could make this a fixture itself:

```python
@pytest.fixture
def test_tree(tmp_path: Path):
    ...
    return tmp_path
...
def test_prunt_oldest_15(self, test_tree: Path)
```

system-tests-new/incremental/test_incremental.py
```diff
@@ -224,3 +228,158 @@ def assert_one_artefact(self, pb_keys, prebuild_groups, prebuild_folder,
...
+
+    # helpers
+
+    def _prune_oldest(self, tmp_path, age):
+        # common code to test pruning by age
+        config = BuildConfig(
+            project_label=PROJECT_LABEL, fab_workspace=tmp_path, multiprocessing=False,
+            steps=[
+                CleanupPrebuilds(older_than=timedelta(days=age))
+            ],
```
MatthewHambley, line 326: You might be able to use a [parameterised test](https://docs.pytest.org/en/7.2.x/how-to/parametrize.html) to save some boilerplate. There may be a problem around interacting with the `test_Tree` fixture mentioned above.

MatthewHambley (changes requested):
Just a few queries and suggestions.

dev_envs/dev_env.yml
```diff
@@ -2,7 +2,7 @@
 name: sci-fab
 dependencies:
   - libclang>=8.0.1
-  - python>=3.8
+  - python==3.7
```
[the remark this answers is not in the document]
bblay: We are, this is just the development environment which we need to force to 3.7.
There are sibling yamls in the same folder for every supported python version, should a dev wish to test on those, and the lowest and highest are used in the rose build-all suite.

source/fab/steps/cleanup_prebuilds.py
bblay: The goal is to support both modes like you say, but also I believe we _do_ handle all the cases you mention.
 - If we provide nothing, then `all_unused` is set
 - If we provide nothing for the first two, then `all_unused` _must_ be set.
 - One or both of the first two is fine.

The only thing we might want to do it make clear that when `all_unused` is set there's no point providing the first two because they are ignored. Perhaps that should raise an error?

source/fab/steps/cleanup_prebuilds.py
bblay: I don't know how to test this properly without something like [TestingBot](https://testingbot.com/), which would be woahverkill. Instead, we could just document the need for a file system which reports access time and drop this. I'm leaning that way now. I think you are too so I'll remove this feature.

system-tests-new/incremental/test_incremental.py
bblay: I tried at the time but didn't find a way. For `test_prune_oldest_15`, the `tmp_path` fixture is passed through to `_prune_oldest` which creates and returns the config and artefacts. I don't understand if the above snippet can help with that?

system-tests-new/incremental/test_incremental.py
bblay: I did a bit of refactoring and was able to parameterise all but one into a single test. It's much more concise, and reads better too now, I think. (Note: I've found it can be awkward to investigate and debug failing parameterised tests, especially when using breakpoints.)

MatthewHambley (approved):
This change looks okay and cleans things up a bit.

PR #154, opened 2022-10-17
ifort & compiler awareness

bblay:
Fixed ifort incompatibilty bug which was introduced by hard-coding `-J` into the compiler to control module output folder.
Fab is now aware of ifort and gfortran flags and applies appropriate flags to control the module folder. Also for `-c` and potentially `-fPIC` / `-fpic`.

 * Replaces the old build_all shell script with a python one. This was originally done because there was logic to build with different compilers, but that doesn't actually work in our dev env as we need different module loads, so now it's just a simple list of calls just like the shell script already was.
 * Removed the optimisation argument from the example gcom build scripts, to reduce clutter and because we weren't using it.
 * There's some new stuff to do with managing flags for known compilers, but we are due a design chat on whether we _ever_ want to do this.
 * Note: Atm build is currently failing for both compilers with 32/64 type mismatch errors, and ifort UM build is failing. We'll need to configure these builds to pass again soon, and automate/cron them ([ticket](https://github.com/metomi/fab/projects/8#card-86322390)).

- [x] merge #147
- [x] update after merge


GiorgiaLine (approved):
Assuming the items marked for discussion in Byron's comments can wait until after the beta release, I think this is god to go.

PR #155, opened 2022-10-18
"Fixup" analysis prebuild paths

bblay:
When we load analysis prebuilds, we know the "local" file for which we want results, but the result file contains the _original_ path. This was causing the compiler to try to compile _the original source_ from the prebuild provider, instead of our local source.
This fix applies a "fixup" to the loaded result path. A new test is included which deletes the original source and fails without the fix.

### DIFF
--- a/system-tests-new/prebuild/test_prebuild.py
+++ b/system-tests-new/prebuild/test_prebuild.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import zlib
 from pathlib import Path
 from unittest import mock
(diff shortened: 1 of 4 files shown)


system-tests-new/prebuild/test_prebuild.py
```diff
@@ -73,26 +75,58 @@ def test_repeatable_fmod_hashes(self, tmp_path):
...
 
         # make sure the exe was built
-        exe = my_config.project_workspace / 'my_prog.exe'
+        exe = second_project.project_workspace / 'my_prog.exe'
         assert exe.exists()
+
+        # make sure the prebuild files are the same
+        for a in file_walk(first_project.prebuild_folder):
+            b = second_project.prebuild_folder / a.name
+            assert files_identical(a, b)
```
MatthewHambley, line 101: This will trap files present in "a" but not in "b". Does it trap files present in "b" but not in "a"?

MatthewHambley (changes requested):
Just one minor issue

MatthewHambley (approved):
Change looks okay for merge

PR #156, opened 2022-10-21
version bump 0.10.0

bblay:


MatthewHambley (approved):
There is nothing controversial in this change. It does highlight that we have the version number in two places which we should fix at some point.

PR #157, opened 2022-10-24
remove the prebuild sharing between projects in the build_all tests

bblay:
Sharing the prebuild folder across projects breaks (at least) atm build and jules rebuild with gfortran, causing a 4 vs 8 byte type mistmatch. This is a cause for concern because the prebuild code isolates the artefacts by hashing the compile flags and more.

I don't think this PR should block the beta release as I['](https://i.pinimg.com/originals/07/94/8f/07948f3fe9076fc809cf5c2f2ab8b1dd.png)m the only one currently running this manual test. However, the shared prebuild problem itself should probably block the release until we understand it.


MatthewHambley (approved):
Understanding this problem is key but making sure it isn't invoked by our tests is a workable stopgap.

PR #158, opened 2022-10-25
dev version bump 1.0.0.dev

bblay:

PR #159, opened 2022-10-25
Remove prebuild flag

bblay:
No way to control the prebuild folder location, it's forced to be inside the project workspace's build_output folder.

I'm still worried that we don't understand what this revealed. It makes me worry that Fab is not properly differentiating incremental/prebuild artefacts by their inputs.

todo:
 - [x] merge #158 before anything else


MatthewHambley (approved):
Glad to see these go as well.

PR #160, opened 2022-10-26
Cron local system tests

bblay:
A cron job for building some local projects, to be run nightly from a dev environment.

We should consider making a ROSE version of this too, with parallel project builds, to speed up manual testing during the day.

### DIFF
--- a/run_configs/_cron/build_all_gfortran.sh
+++ b/run_configs/_cron/build_all_gfortran.sh
@@ -1,0 +1,10 @@
+set -e
+
+# set up the MO gfortran environment with Fab in conda
+module use /data/users/lfric/modules/modulefiles.rhel7
+module load environment/lfric/gnu
+source /opt/conda/etc/profile.d/conda.sh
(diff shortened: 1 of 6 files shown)


run_configs/_cron/build_all_gfortran.sh
```diff
@@ -0,0 +1,10 @@
+set -e
```
MatthewHambley, line 1: I also like to set `-u` which treats dereferencing an undefined macro as an error. This is probably not relevant here as no macros are used. Just thought I'd mention it though.

MatthewHambley (changes requested):
This looks fine, just one comment.

run_configs/_cron/build_all_gfortran.sh
bblay: Thanks, I didn't know about that!

MatthewHambley (approved):
Happy with this.

PR #162, opened 2022-11-02
AnalysedFile automatic file hash

### LINKED ISSUE #161: AnalysisFile automatic hash
When providing a workaround for an fparser problem, the user might have to provide their own `AnalysedFile` object. They should not have to provide a dummy hash value. Fab should calculate if for them.

bblay:
When the user is forced to provide an `AnlsysedFile` to workaround an fparser problem, they should not have to provide a file hash. Also:
- moved `get_mod_hashes()` (and it's unit test) out of util and into _compile_fortran.py_, where it belongs, to stop a circular import.

Note: `AnalysedFile` will now calculate the file hash with _lazy evaluation_ if it's not provided. We would trigger a call to `AnalysedFile.__hash__` by putting the object in a set, for example, which would then trigger the code to calculate the file hash. However, the target file might not exist at this time since we're _defining_ the steps, not running them so there's been no grab. (Worse, a previous version of the file might exist.) To discourage the user from putting these objects in a set, the `Analyse` step now wants a list of these objects instead.

There's still a small risk here, as the user _could_ still put them in a set. However, a number of things would have to happen, seemingly posing a low risk(?): a) doing an fparser workaround, b) not providing a hash, c) passing a set not a list, d) having the grab step as part of the build config, instead of in it's own grab config run beforehand. Thoughts welcome!

Resolves #161.

Todo:
 - [ ] Discuss: Might be better to keep `AnalysedFile` out of the user's picture altogether and just provide a helper class specifically for specifying fparser workarounds. Let Fab deal with _"oh it can't go in a set because hashing"_, behind the scenes, don't expose the user to that.


MatthewHambley (approved):
Lazy hashing may offer a performance improvement on top of fixing the immediate problem.

A convenience function for creating exceptions and thereby insulating the user from internal workings is good for everyone. I commend it for a future ticket..

Issue #164, opened 2022-11-04
very long errors


bblay:
Make these error message more readable by printing the newlines as such and wrapping remaining long lines.

```
Traceback (most recent call last):
  File "/home/h01/jhickson/.conda/envs/sci-fab/lib/python3.10/site-packages/fab/build_config.py", line 96, in run
    step.run(artefact_store=artefact_store, config=self)
  File "/home/h01/jhickson/.conda/envs/sci-fab/lib/python3.10/site-packages/fab/steps/compile_fortran.py", line 63, in run
    check_for_errors(results_this_pass, caller_label=self.name)
  File "/home/h01/jhickson/.conda/envs/sci-fab/lib/python3.10/site-packages/fab/util.py", line 163, in check_for_errors
    raise RuntimeError(
[...]

PR #165, opened 2022-11-07
Error output formatting

bblay:
Insert a newline before tool output in error messages, for #164.  The user was also seeing newlines _within_ the error messages being printed as `\\n` but that has already been addressed.

Also: The new test only has one step, which deliberately fails, so there's never any step timings in the metrics. This PR includes a change which stops that from causing a key error.

todo:
 - [x] #168 then update this
 - [x] #169 then backport this

### DIFF
--- a/source/fab/metrics.py
+++ b/source/fab/metrics.py
@@ -242,14 +242,18 @@ def metrics_summary(metrics_folder: Path):
     run = metrics['run']
     time_taken = datetime.timedelta(seconds=int(run['time taken']))
     min_label_thresh = time_taken.seconds * 0.01
-    step_metrics = metrics['steps'].items()
-    step_times = [kv[1] for kv in step_metrics]
-    step_labels = [kv[0] if kv[1] > min_label_thresh else "" for kv in step_metrics]
+    step_totals = metrics.get('steps')
+    if step_totals:
+        step_metrics = step_totals.items()
+        step_times = [kv[1] for kv in step_metrics]
+        step_labels = [kv[0] if kv[1] > min_label_thresh else "" for kv in step_metrics]
 
-    plt.pie(step_times, labels=step_labels, normalize=True,
-            wedgeprops={"linewidth": 1, "edgecolor": "white"})
-    plt.suptitle(f"{run['label']} took {time_taken}\n"
-                 f"on {run['sysname']}, {run['nodename']}, {run['machine']}")
-    plt.figtext(0.99, 0.01, f"{metrics['run']['datetime']}", horizontalalignment='right', fontsize='x-small')
-    plt.savefig(metrics_folder / "pie.png")
-    plt.close()
+        plt.pie(step_times, labels=step_labels, normalize=True,
+                wedgeprops={"linewidth": 1, "edgecolor": "white"})
+        plt.suptitle(f"{run['label']} took {time_taken}\n"
+                     f"on {run['sysname']}, {run['nodename']}, {run['machine']}")
+        plt.figtext(0.99, 0.01, f"{metrics['run']['datetime']}", horizontalalignment='right', fontsize='x-small')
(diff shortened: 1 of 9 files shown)


bblay:
Running the new unit test in v0.9.0a1 we can see double backslashes in the output.

```
tests/test_build_config.py:16 (TestBuildConfig.test_error_newlines)
'1\n2\n3' != ('\n'
 '\n'
 'Error running build steps:\n'
 "('Error calling compiler:', RuntimeError('foo error\\n1\\n2\\n3'))\n"
 '\n'
 '1 error(s) found during compile fortran')

```

tests/unit_tests/test_build_config.py
```diff
@@ -0,0 +1,36 @@
...
+from unittest import mock
+
+from fab.build_config import BuildConfig
+from fab.dep_tree import AnalysedFile
+from fab.steps.compile_fortran import CompileFortran
+
+
+class TestBuildConfig(object):
+
+    def test_error_newlines(self, tmp_path):
```
MatthewHambley, line 16: Is the purpose of this test to ensure that newlines are preserved through the error reporting process. It's not clear to me how much of what is being tested.

source/fab/steps/compile_c.py
```diff
@@ -114,7 +114,7 @@ def _compile_file(self, analysed_file: AnalysedFile):
                 try:
                     run_command(command)
                 except Exception as err:
-                    return TaskException(f"error compiling {analysed_file.fpath}: {err}")
+                    return TaskException(f"error compiling {analysed_file.fpath}:\n{err}")
```
MatthewHambley, line 117: Might it be worth creating our own exception class which takes a message, file and exception on its constructor? Then its `__str__` method could do this formatting across the board.

source/fab/metrics.py
```diff
@@ -242,14 +242,18 @@ def metrics_summary(metrics_folder: Path):
...
-    step_labels = [kv[0] if kv[1] > min_label_thresh else "" for kv in step_metrics]
-
-    plt.pie(step_times, labels=step_labels, normalize=True,
-            wedgeprops={"linewidth": 1, "edgecolor": "white"})
-    plt.suptitle(f"{run['label']} took {time_taken}\n"
-                 f"on {run['sysname']}, {run['nodename']}, {run['machine']}")
-    plt.figtext(0.99, 0.01, f"{metrics['run']['datetime']}", horizontalalignment='right', fontsize='x-small')
-    plt.savefig(metrics_folder / "pie.png")
-    plt.close()
+    step_totals = metrics.get('steps')
```
MatthewHambley, line 245: Is this change mentioned in the summary? What's it about?

MatthewHambley (changes requested):
A few questions about this change.

tests/unit_tests/test_build_config.py
bblay: Yes that's right. When there's an error containing new lines, v0.9.0a1 does not display the newlines, it prints `\\n`. I've tried to make that more clear in the test's comment.

source/fab/steps/compile_c.py
bblay: Yes, I think there's likely a few useful Fab exceptions we could make. There's a ticket to review the exception handling. I'd suggest that would fall under it's remit?

source/fab/metrics.py
bblay: It's a bit of robustness. The new test only has one step, which deliberately fails, so there's never any step timings in the metrics. This change stops that from causing a key error. I'll add this to the description too.

MatthewHambley: You never got round to it but we have a backlog of changes so maybe for a future ticket?

MatthewHambley (approved):
I'm approving this as is on the understanding that the missing comment will be added to a subsequent change.

PR #173, opened 2022-11-09
stop calling integration tests

bblay:
Thought this was already removed.

There are currently no integration tests, since we stopped using csv to store analysis and fortran compilation results. There's a dummy test because we're still trying to call them, which I don't think we can remove until this is merged.


matthewrmshin (approved):
Can't argue with this one.

PR #175, opened 2022-11-10
0.10 Mp hang bug

bblay:
Turn off multiprocessing when analysing C from Python 3.7.


MatthewHambley (approved):
This is the same fix recently sent to trunk.

PR #176, opened 2022-11-10
Mp hang bug

bblay:
Turn off multiprocessing when analysing C from Python 3.7.


MatthewHambley (approved):
This looks to do the necessary.

PR #177, opened 2022-11-11
185 Rose suite system tests

### LINKED ISSUE #185: nightly crons -> rose
Turn the nightly cron system tests into a rose suite. Builds could be run in parallel. The suite could be run at any time without blocking a dev.

bblay:
I'm losing a fair bit of time waiting for the offline system tests to run on the VDI. This rose suite runs them all in parallel. It's quicker ~and I can work while they run~.

This suite runs all the MO projects we regularly build with Fab. It builds them with Python versions 3.7 and 3.11, and it builds them with both gfortran and ifort.

This method of running the offline system tests differs from the nightly cron in that it runs the currently checked out branch, not the head of trunk.

todo:
 - [x] check they all build ok
    - (except for the couple of ifort ones we haven't configured properly yet)

Resolves #185

### DIFF
--- a/run_configs/_rose_all/suite.rc
+++ b/run_configs/_rose_all/suite.rc
@@ -1,89 +1,84 @@
+[cylc]
+    [[parameters]]
+        pyver = "7", "11"  # we can't put a full stop in params, like "3.7", because it causes invalid job names
+        compiler = gnu, ifort
+
 [scheduling]
     [[dependencies]]
-        {% for PYTHON_VERSION in PYTHON_VERSIONS %}
-            {% for COMPILER in COMPILERS %}
-                graph = """
-                    grab_gcom_py{{ PYTHON_VERSION }}_{{ COMPILER }} => build_gcom_ar_py{{ PYTHON_VERSION }}_{{ COMPILER }}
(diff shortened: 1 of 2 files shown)


run_configs/_rose_all/suite.rc
```diff
@@ -0,0 +1,89 @@
+[scheduling]
+    [[dependencies]]
+        {% for PYTHON_VERSION in PYTHON_VERSIONS %}
+            {% for COMPILER in COMPILERS %}
```
MatthewHambley, line 4: Rather than using Jinja2 looping you might find that [parameterised tasks](https://cylc.github.io/cylc-doc/7.8.8/html/suite-config.html#parameterized-tasks) give cleaner syntax.

MatthewHambley (changes requested):
A simple suite with an opportunity to use some built-in syntax.

run_configs/_rose_all/suite.rc
bblay: Oh that's so much better, thank you.

MatthewHambley (approved):
Basic testing ready to go

PR #178, opened 2022-11-11
174 incremental c

### LINKED ISSUE #174: incremental c
todo:
 - [ ] coordinate with #153 to register incremental c artefacts as "current"

bblay:
Incremental C building.
- Function to get C compiler largely the same as for Fortran, so now in `util.get_tool()`.
- Added `get_fortran_compiler()` for user convenience.
- Added `-c` if missing, with warning. This seems like a step towards C compiler awareness, like we have with Fortran.
- Moved `get_compiler_version()` to util because C compiler version format matches Fortran.

todo:
 - [ ] coordinate with #153 to register incremental c artefacts as "current"

Closes #174

### DIFF
--- a/source/fab/steps/compile_c.py
+++ b/source/fab/steps/compile_c.py
@@ -56,7 +56,7 @@ class CompileC(Step):
         """
         super().__init__(name=name)
 
-        self.compiler, compiler_flags = get_tool(compiler or os.getenv('CC', 'gcc -c'))
+        self.compiler, compiler_flags = get_tool(compiler or os.getenv('CC', 'gcc -c'))  # type: ignore
         self.compiler_version = get_compiler_version(self.compiler)
         logger.info(f'c compiler is {self.compiler} {self.compiler_version}')
 
--- a/source/fab/steps/compile_c.py
+++ b/source/fab/steps/compile_c.py
@@ -20,7 +20,8 @@ from fab.dep_tree import AnalysedFile
 from fab.metrics import send_metric
 from fab.steps import check_for_errors, Step
 from fab.tasks import TaskException
-from fab.util import CompiledFile, run_command, log_or_dot, Timer, by_type, flags_checksum
+from fab.util import CompiledFile, run_command, log_or_dot, Timer, by_type, flags_checksum, get_compiler_version, \
+    get_tool
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


source/fab/steps/compile_fortran.py
```diff
@@ -218,8 +214,6 @@ def process_file(self, analysed_file: AnalysedFile):
             Before compiling a file, we calculate the combo hashes and see if the output files already exists.
 
         """
-        # todo: include compiler version in hashes
```
MatthewHambley, line 221: Has this been done? `_get_obj_combo_hash` doesn't seem to do it. Is it no longer relevant?

source/fab/steps/compile_c.py
```diff
@@ -56,7 +57,19 @@ def __init__(self, compiler: Optional[str] = None, common_flags: Optional[List[s
         """
         super().__init__(name=name)
 
-        self.compiler = compiler or os.getenv('CC', 'gcc -c')
+        self.compiler, compiler_flags = get_tool(compiler or os.getenv('CC', 'gcc -c'))  # type: ignore
```
MatthewHambley, line 60: Should there be a `get_c_compiler` to mirror the `get_fortran_compiler`? Or should the latter be replaced by `get_tool`?

MatthewHambley (changes requested):
A few questions regarding the extent of this change.

source/fab/steps/compile_fortran.py
bblay: Yes, done and merged in #150. It's on line 267, in `_get_obj_combo_hash`.

source/fab/steps/compile_c.py
bblay: We could definitely remove `get_fortran_compiler` and just call `get_tool(compiler or os.getenv('FC', ''))`. In fact this PR did that at one point. The reason I added `get_fortran_compiler` was as a convenience function for user configs. I think it improves ease and readability in user code. Happy to remove if preferred, of course.

source/fab/steps/compile_fortran.py
bblay: (now on line 145, post merge)

source/fab/steps/compile_c.py
MatthewHambley: At some point people will start to use C++ modules. That will precipitate exactly the same issues surrounding Fortran modules. We will then need per-compiler builds. In the meantime we can leave it as is.

MatthewHambley (approved):
I am happy that we know where the limitations are.

PR #181, opened 2022-11-14
180 disable c without clang

### LINKED ISSUE #180: conditional c support
handle clang problem, disable c support

bblay:
If clang is not available, don't attempt to do anything in the c analyser code. Warn but don't fail.

Note: We could raise an exception but we've already seen config sharing which includes c steps when there's no c code. In this case  it would seem better to just notify but continue.

Resolves #180

### DIFF
--- a/source/fab/tasks/c.py
+++ b/source/fab/tasks/c.py
@@ -6,11 +6,16 @@ C language handling classes.
 
 """
 import logging
+import warnings
 from collections import deque
 from pathlib import Path
 from typing import List, Optional
 
-import clang.cindex  # type: ignore
+try:
+    import clang  # type: ignore
+    import clang.cindex  # type: ignore
+except ImportError:
+    clang = None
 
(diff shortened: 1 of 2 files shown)


source/fab/tasks/c.py
```diff
@@ -83,6 +88,12 @@ def _check_for_include(self, lineno) -> Optional[str]:
             return None
 
     def run(self, fpath: Path):
+
+        if not clang:
+            msg = 'clang not available, C analysis disabled'
+            warnings.warn(msg, ImportWarning)
+            return ImportWarning(msg)
```
bblay, line 95: Note to reviewer: We use a multiprocessing pattern where child processes return either a successfully created object, or an error. Errors are then gathered and reported together. There's also a ticket on the [development board](https://github.com/orgs/metomi/projects/1/views/1) to review exception handling.

matthewrmshin (approved):

PR #182, opened 2022-11-14
update old system tests

bblay:
update old system tests
- CUseHeader
- FortranPreProcess
- MinimalC
- MinimalFortran


MatthewHambley (approved):
This is a fairly simple relocation of the test cases, reimplemented for the current version. I checked that they are actually being run as part of the GitHub workflow and all seems will.

PR #188, opened 2022-11-16
189 example run configs: ifort

### LINKED ISSUE #189: ifort build all
Finishing touch: I think we just need to find the type specification for ifort, like the ones we have for gfrotran, to avoid the type mismatch errors. I think this will let us build UM and LFRic ATM with ifort.

This is really an issue with the nightly run configs, not Fab itself.

done:
 - replaced the hard-coded gfortran -J flag
 - both gfortran and ifort are tested regularly in #160

bblay:
Get the last couple of example build configs worting with ifort.

- The example UM build config now works with ifort.
- (todo: atm)

resolves #189

### DIFF
--- a/run_configs/um/build_um.py
+++ b/run_configs/um/build_um.py
@@ -38,7 +38,7 @@ logger = logging.getLogger('fab')
 # todo: fail fast, check gcom exists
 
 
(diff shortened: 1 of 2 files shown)


bblay:
got stuck on atm, pushing um fix separately

PR #191, opened 2022-11-17
Ifort um

bblay:
UM example config now works with ifort.

 - Removed the opt argument from our example um script because I think it differs between compilers and doesn't seem helpful right now.
 - Addresses part of #189 (i'll need help with the atm config, in a separate pr).
 - Moved a couple of steps' log lines from `__init__` to `run` because they were only going to the screen (the file logging handler is set up in run, and there's also [a ticket](https://github.com/orgs/metomi/projects/1/views/1?filterQuery=logg) to revisit the logging).

### DIFF
--- a/run_configs/build_all.py
+++ b/run_configs/build_all.py
@@ -35,7 +35,7 @@ def build_all():
     ]
 
     # skip these for now, until we configure them to build again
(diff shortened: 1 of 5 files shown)


matthewrmshin (approved):
I'll leave it to the system team to decide how to name and configure the optimisation levels of different platforms/architectures/compilers.

PR #197, opened 2022-11-28
195 grab git

bblay:
New `GrabGit` grab step. Allows both shallow and deep grab, the latter to support grabbing a commit hash.
- Includes some minor tidying:
   - Moved some common code into the base class.
   - Existence of the `GrabSvn` class is no longer gated behind svn import, just access to it's functionality. This means better error reporting plus build scripts won't fail to import (in case you want to reuse something). It's a better pattern I was exploring whilst planning to use a git library, which I ended up not doing (see below).
- There's a few mypy ignore directives that we probably shouldn't be happy about. It's not clear what to do about them though.

~Question: Would we prefer to use a python git library instead of whatever's installed in the environment? Consistency vs extra dependencies. Ideally we'd have both a cli git version to reduce dependencies and a git library version to reduce dependency on user environment and provide consistency across platforms.~ `GitPython` uses the installed cli tool.

Todo:
 - [ ] switch to git library, as discussed offline


bblay:
Closing for the moment, I've realised that _all_ repo grabbers must probably insist on having a destination label (i.e sub folder) because they all need to keep their _.git_, _.svn_, etc, folders separate. Only the folder grab can omit the destination label and put stuff all in the source root.  I'll make that change in here so the svn and fcm upgrades can base off that.

bblay:
Closing while I make the system tests more thorough.

bblay:
Good job, as there was one case where they failed.

source/fab/steps/grab.py
```diff
@@ -28,25 +32,33 @@
...
 
         """
-        super().__init__(name=name or f'{self.__class__.__name__} {dst or src}'.strip())
+        super().__init__(name=name or f'{self.__class__.__name__} {src} {revision}'.strip())
         self.src: str = src
         self.dst_label: str = dst or ''
+        self.revision = revision
+
+        # runtime
+        self._dst: Optional[Path] = None
```
MatthewHambley, line 61: Your typing problems may well be down to the fact that this member is declared `Optional` but it is not fenced with `if self._dst is not None` when used.

However there is a pungent code smell here. It looks like you're using a nominally private member variable to communicate between child and parent classes.

At the very least some sort of accessor function should be used rather than direct access. A `get_destination` method or a `@property` method.

MatthewHambley (changes requested):
This looks like a good basis for any additional Git support we may need.

source/fab/build_config.py
```diff
@@ -43,8 +43,15 @@ def __init__(self, project_label: str, source_root: Optional[Path] = None, steps
...
-            This is useful, for example, when the :py:mod:`fab.steps.grab <grab>` is in a separate script to be run
-            less frequently. In this scenario, the source code will be found in a different project workspace folder.
+            This is useful, for example, when the :py:mod:`~fab.steps.grab` is in a separate script to be run
+            less frequently. In this scenario, the source code might be found in a different project workspace folder.
+            This argument is intended to be used *instead* of grab steps.
+
+            .. note::
+
+                **Take care if using this argument with grab steps**
+                because you could write to a source folder *outside* your project workspace.
```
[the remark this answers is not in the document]
bblay: Good question. Not yet. Some options:
- We could raise an exception if using a grab step when the source folder is overridden. I don't think we've got a use case for using both at the same time; they seem mutually exclusive so that seems safe to do.
- Remove the source folder override. We can just use a folder grab instead and the problem goes away. We added the override during early development when porting FCM build configs. I don't think it's essential.

source/fab/steps/grab.py
bblay: I'd argue that there's no such thing as communication between a parent and child class but that's moot because the property suggestion is nicer anyway!

bblay: Ah, I couldn't do a property because it depends on another runtime attribute, the config.

Interestingly, that config is also made available using the same _"runtime attribute pattern"_ (made up term) via `self._config` so child processes can read it.

The fortran compiler also uses two runtime-only attributes called `self._stage` and `self._mod_hashes`, and the two language parsers both have one.

Perhaps we should have an OO chat about this?

source/fab/build_config.py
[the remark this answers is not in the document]
MatthewHambley: If this problem exists purely because of unneeded functionality then the solution seems clear. Remove the cause of the problem. It may be necessary to document the use of the file system grab in this situation if it is going to be common practice.

run_configs/tiny_fortran/build_tiny_fortran.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # We want a separate project folder for each compiler. Find out which compiler we'll be using.
+    compiler, _ = get_fortran_compiler(compiler)
+    config = BuildConfig(project_label=f'tiny_fortran {compiler}')
+
+    logger.info(f'building tiny fortran {config.project_label}')
+
+    config.steps = [
+
+        GrabGit(src='https://github.com/bblay/tiny_fortran.git', revision='main', dst='src'),
```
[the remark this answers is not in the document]
bblay: Yes, but then we won't pick up failures which are specific to remote repos. This test _was failing_ while the tests which use the archive were passing. That was because I was using the library incorrectly, so this remote repo test is the only thing stopping us breaking the main use case for this code without knowing.

source/fab/steps/grab.py
MatthewHambley: We should definitely have a discussion about this but probably not in this change. I've raised #210 as a venue.

MatthewHambley (changes requested):
I've lost track of where this review is but I've added some comments.

MatthewHambley (approved):
Lets go with this as our base Git implementation.

PR #172, opened 2022-11-09
171 cron creates conda

bblay:
nightly cron system tests now create a fresh conda and install the head of Fab, instead of activating the developer's conda with whatever branch they happen to have checked out.

Tested locally.
Resolves #171.

PR #198, opened 2022-11-29
fix cron conda yaml folder

bblay:
The conda yaml has moved but the cron is still looking in the old place.

Tested locally.
(I tested #172 locally too. I would have expected that to have failed without this change...)


bblay:
Closing because
image

PR #199, opened 2022-11-29
increase api ref toc depth to show steps

bblay:
There wasn't an easy, obvious way to see a list of steps in the docs.
Increasing the apidoc depth from 3 to 5 means the steps are visible from the [api reference root page](https://metomi.github.io/fab/apidoc/modules.html).

There's a difference between local and [github build](https://bblay.github.io/fab/) (sphinx v4.3.2 and v5.3.0 respectively). The latter has all the member classes and their methods, which is quite noisy, but at least the information is all there. It seems to ignore the `--module-first` option, too. The difference might be due to having versions, or related to `SPHINX_APIDOC_OPTIONS`.

locally:

gh action:


GiorgiaLine (approved):
This looks fine to me, approved for merge.

GiorgiaLine (approved):
Re-approving after conflicts were resolved.

PR #204, opened 2023-01-18
fix automatic cleanup step addition

bblay:
The automatic addition of a cleanup step was not tested(!), and broke.


MatthewHambley (approved):
Simple bug fix. We might need to look at the creation of temporary lists in the future but we are not running out of memory yet.

PR #205, opened 2023-01-19
Don't filewalk prebuild

bblay:
During the incremental psyclone work, it was found that source code discovery using `file_walk()` was traversing into the prebuild folder. This became a problem when we started storing prebuilds of generated fortran, which would then be picked up and analysed in a subsequent run. We don't want that to happen. We want to pull what we need back out from the prebuild folder.

This PR adds an option to `file_walk()` to ignore given folders.


tests/unit_tests/test_util.py
```diff
@@ -51,3 +51,23 @@ def test_constructor_suffix_vector(self):
...
+        f = tmp_path / 'foo/bar/foo.txt'
+        pbf = tmp_path / 'foo/bar/_prebuild/foo.txt'
+
+        pbf.parent.mkdir(parents=True)
+        f.touch()
+        pbf.touch()
+
+        return f, pbf
+
+    def test_ignore(self, files, tmp_path):
```
MatthewHambley, line 69: I'm not sure this is doing the right thing. Doesn't `tmp_path` used here and `tmp_path` used by `files` generate different paths?

MatthewHambley (changes requested):
A quick question about fixturing.

tests/unit_tests/test_util.py
bblay: Good question but no, it's set per test invocation. In this case, printing from both prints `/var/tmp/pytest-of-bblay/pytest-1356/test_ignore0` twice.

MatthewHambley (approved):
Simple solution to the problem. We need an issue to investigate why it's visiting prebuilds at all.

PR #206, opened 2023-01-22
Picasso

bblay:
This PR allows us to manually trigger a Picasso build, giving us a singularity container with Fab and dependencies. It's based on a recent version of Ubuntu with a newer glibc than on the VDI.

Follow up work might branch out into multiple images:
 - Different versions of Python for testing, which is currently not done in a sustainable way.
 - A dev environment.
 - Fab versions released in their own container.
 - Something with LLVM to resolve a problem Matthew is more able to describe.

We probably don't want to trigger these builds from push or prs because a) we probably only need to build infrequently and b) due to the cross-repo nature of the action we'd need to give fab forks (i.e anyone) source permissions to the source repo. A disadvantage of this that I don't think we can quickly iterate with PRs to tweak the Picasso config, we have to wait until it's merged before we can see if it works.


matthewrmshin (commented):
Looks good, but recommend that you use the `python3` command instead of the `python` command. Some newer Linux distro versions have chosen to only provide a `python3` command and not a `python` command.

matthewrmshin (approved):
Looks good to me.

PR #208, opened 2023-01-25
grab subpackage (no change)

bblay:
Turns the grab module into a package.
This PR contains no code changes other than moving classes into the grab sub package.

It paves the way for the fcm merge branch.
We don't want all these diffs to pollute that PR.


MatthewHambley (approved):
As noted, this is a simple re-organisation which generates a lot of apparent differences.

PR #207, opened 2023-01-25
Grab archive

bblay:
todo:
 - [x] merge #197, then uncompress the archive it contains as a test for this pr

PR #209, opened 2023-01-26
Fcm merge

bblay:
Upgrade Fcm capabilities. Separate steps for:
 - export
 - checkout/update
 - merge

todo:
 - merge #207, which contains much of these diffs in a non-functions change PR, thereby reducing the size of this pr


source/fab/steps/grab/fcm.py
```diff
@@ -3,23 +3,16 @@
...
 #  which you should have received as part of this distribution
 # ##############################################################################
-from typing import Dict
+from fab.steps.grab.svn import SvnCheckout, SvnExport, SvnMerge
 
-from fab.steps.grab import GrabSourceBase
-from fab.tools import run_command
 
+class FcmExport(SvnExport):
+    command = 'fcm'
```
MatthewHambley, line 10: I don't like this breaking of encapsulation. I would rather have the command passed to the constructor so it is explicit and encapsulation is respected.

i.e. The base Subversion class takes the subversion command as an argument on its constructor. The FCM constructor then calles `super().__init__('fcm')`.

source/fab/steps/grab/svn.py
```diff
@@ -3,32 +3,160 @@
...
+        if not self._dst.exists():  # type: ignore
+            run_command([
+                self.command, 'checkout',
+                *self._cli_revision_parts(),
+                self.src, str(self._dst)
+            ])
+
+        else:
+            # working copy?
+            if self._is_working_copy(self._dst):  # type: ignore
```
MatthewHambley, line 131: If `_is_working_copy(...)` is called externally (including from a child) then it isn't "private".

source/fab/steps/grab/svn.py
```diff
@@ -3,32 +3,160 @@
...
+                # update
+                # todo: ensure the existing checkout is from self.src?
+                run_command([self.command, 'update', *self._cli_revision_parts()], cwd=self._dst)  # type: ignore
+            else:
+                # we can't deal with an existing folder that isn't a working copy
+                raise ValueError(f"destination exists but is not an fcm working copy: '{self._dst}'")
+
+
+class SvnMerge(GrabSvnBase):
```
MatthewHambley, line 140: Be aware that the major point of divergance between Subversion and FCM is in the merge code. FCM implements its own merge algorithm and doesn't use Subversion's. As such it may not be a good idea to try and support FCM and Subversion with the same method.

source/fab/steps/grab/svn.py
```diff
@@ -3,32 +3,160 @@
...
+            rev_url = f'{self.src}'
+            if self.revision is not None:
+                rev_url += f'@{self.revision}'
+
+            res = run_command([self.command, 'merge', '--non-interactive', rev_url], cwd=self._dst)
 
-        r = svn.remote.RemoteClient(self.src)
-        r.export(str(self._dst), revision=self.revision, force=True)
+            # Fcm doesn't return an error code when there's a conflict, so we have to scan the output.
+            if 'Summary of conflicts:' in res:
```
MatthewHambley, line 161: Rather than scraping strings which may change is there a more reliable approach? Can `fcm status` be used?

system-tests-new/svn_fcm/test_svn_fcm.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Run the test twice, once with SvnExport and once with FcmExport - depending on which tools are available.
+    @pytest.mark.parametrize('export_class', export_classes)
+    def test_export(self, repo_url, config, export_class):
+        export = export_class(src=f'{repo_url}/proj/main/trunk', dst='proj')
+        export.run(artefact_store=None, config=config)
+
+        # Make sure we can export twice.
+        # Todo: should the export step wipe the destination first?
+        export.run(artefact_store=None, config=config)
```
MatthewHambley, line 53: There are no assertions in this test so it makes sure the calls return but I don't think they can know if anything has been achieved.

system-tests-new/svn_fcm/test_svn_fcm.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class TestFcmCheckout(object):
+
+    @pytest.mark.parametrize('checkout_class', checkout_classes)
+    def test_new_folder(self, repo_url, config, checkout_class):
+        checkout = checkout_class(src=f'{repo_url}/proj/main/trunk', dst='proj')
+        checkout.run(artefact_store=None, config=config)
+
+        file2_txt = open(config.source_root / 'proj/file2.txt').read()
```
MatthewHambley, line 63: There is a `read_text` method on `Path` objects which can potentially aid readability.

system-tests-new/svn_fcm/test_svn_fcm.py
```diff
@@ -0,0 +1,191 @@
...
+        assert "This is sentence two in file two." in file2_txt
```
MatthewHambley, line 64: Might use `str.endswith()` to ensure the string is in the right place as well as present.

tests/unit_tests/steps/test_grab.py
```diff
@@ -42,41 +43,40 @@ def test_no_revision(self):
...
 
         mock_config = SimpleNamespace(source_root=source_root)
         with mock.patch('pathlib.Path.mkdir'):
-            with mock.patch('fab.steps.grab.fcm.run_command') as mock_run:
+            with mock.patch('fab.steps.grab.svn.run_command') as mock_run:
                 grabber.run(artefact_store={}, config=mock_config)
 
-        mock_run.assert_called_once_with(['fcm', 'export', '--force', source_url, str(source_root / dst_label)])
+        mock_run.assert_has_calls([
+            call(['fcm', 'help']),
```
MatthewHambley, line 54: Is this left over debug code or is there a reason for it?

tests/unit_tests/steps/test_grab.py
```diff
@@ -42,41 +43,40 @@ def test_no_revision(self):
...
         mock_config = SimpleNamespace(source_root=source_root)
         with mock.patch('pathlib.Path.mkdir'):
-            with mock.patch('fab.steps.grab.fcm.run_command') as mock_run:
+            with mock.patch('fab.steps.grab.svn.run_command') as mock_run:
                 grabber.run(artefact_store={}, config=mock_config)
 
-        mock_run.assert_called_once_with(
-            ['fcm', 'export', '--force', f'{source_url}@{revision}', str(source_root / dst_label)])
+        mock_run.assert_has_calls([
+            call(['fcm', 'help']),
```
MatthewHambley, line 71: It's that call to "help" again.

source/fab/steps/grab/fcm.py
bblay: Absolutely we can do that. There's a point about build script readability but we can address that with helper functions instead.

source/fab/steps/grab/svn.py
bblay: In Python, a _double_ preceding underscore indicates private members and invokes name mangling. A _single_ preceding underscore indicates protected members, does not invoke name mangling and is accessible from subclasses.

This code:
```
class Base(object):
    def __private_func(self):
        print("subclasses can't call me")
    def _protected_func(self):
        print("subclasses can call me")

class Derrived(Base):
    pass

if __name__ == '__main__':
    Derrived()._protected_func()
    Derrived().__private_func()
```

produces this output:
```
subclasses can call me
Traceback (most recent call last):
  File "/home/h02/bblay/git/fab/temp/private_vs_protected.py", line 17, in <module>
    Derrived().__private_func()
AttributeError: 'Derrived' object has no attribute '__private_func'. Did you mean: '_Base__private_func'?
```

source/fab/steps/grab/svn.py
bblay: I agree, I don't like the dependency on another tool's output formatting. I didn't see an obvious better way. I'll raise a ticket to record this for future attention.

source/fab/steps/grab/svn.py
bblay: Hmm, there's a risk the user will try to cross-merge. I'm not sure that risk comes from sharing common code though. I think it's a configuration/user error because they have to specify either svn or fcm. I wonder if we can invoke the given tool to see if it's an svn or fcm working copy. Thinking ahead though, I understand we will _want_ to be able to cross merge.

system-tests-new/svn_fcm/test_svn_fcm.py
bblay: Agreed, it could be more meticulous. I'll update it to export two different revisions and verify them both.

bblay: ^ same applies to a few others too.

system-tests-new/svn_fcm/test_svn_fcm.py
bblay: That's nicer, thanks.

tests/unit_tests/steps/test_grab.py
bblay: No, it's a replacement for a little check that used to be done in the module upon import, which checked if the tool was available on the command line. After unifying svn and fcm, the check moved into the run method. It's not expensive but if we don't want to run this more than once we change it, if you prefer?

source/fab/steps/grab/svn.py
MatthewHambley: I did not know about the double underscore thing although upon researching it I see it was introduced in 2001 with PEP8. Unfortunately this doesn't change the problem here as it is rarely a good idea to use protected. In fact the developers of C++ consider its existence to be their biggest blunder.

If a child needs to access a member variable then it should be through an accessor (property is fine) and never directly regardless of whether it is possible.

Directly accessing parent variables means you have to know what is in the parent and what consistency requirements it has. Using an accessor means you only need to know the API and consistency is maintained for free.

source/fab/steps/grab/svn.py
MatthewHambley: There are a lot of problems involved in this regardless of what we do. My immediate concern is that `fcm merge` not only behaves differently to `svn merge` but it also takes different arguments. This is not a problem at the moment but may become one later. I propose we stick with what you have for the moment but be aware that it may need t change later.

source/fab/steps/grab/svn.py
MatthewHambley: Subversion offers an `-xml` argument which produces a well defined output you can rely on. I don't think FCM does though.

MatthewHambley (changes requested):
W're in the weeds of OO design now.

bblay:
> I'm sorry, you are quite right about it being a class variable rather than an member variable. Unfortunately that makes it even worse.
> 
> It means that in any build script which has both a Subversion and an FCM grabber we will have no idea which tool is being used. They will both be using the same tool and it will likely depend on which was instantiated most recently.

Are you sure class variables work like that?

A class variable is evaluated _on declaration_, before any instantiation takes place, when the class itself is being evaluated. It's shared between all instances.

It's easy to check if it behaves the way you're suggesting:

```
from fab.steps.grab.fcm import FcmCheckout
from fab.steps.grab.svn import SvnCheckout

f1 = FcmCheckout('', '')
s1 = SvnCheckout('', '')

f2 = FcmCheckout('', '')
s2 = SvnCheckout('', '')

print(f1.command)
print(f2.command)
print(s1.command)
print(s2.command)
print()
print(f1.command)
print(s1.command)
print(f2.command)
print(s2.command)
```
```
fcm
fcm
svn
svn

fcm
svn
fcm
svn
```
We easily know exactly which command is being called by each class. There's no ambiguity, no race condition. This is how class variables work, unless I'm having a huge brain fart - entire possible!

Calling a constructor won't change the value of a class variable, even by accident:
```
class C(object):
    att = 1
    def __init__(self):
        self.att = 2

c = C()
print(c.att)
print(c.__class__.att)
```
```
2
1
```

I'm happy to switch to dependency injection for the command, no problem at all.
But if the request is potentially based on a misunderstanding of Python, I think it's also useful for us to explore that, just to be sure.

bblay:
> I did not know about the double underscore thing although upon researching it I see it was introduced in 2001 with PEP8. Unfortunately this doesn't change the problem here as it is rarely a good idea to use protected. In fact the developers of C++ consider its existence to be their biggest blunder.
> 
> If a child needs to access a member variable then it should be through an accessor (property is fine) and never directly regardless of whether it is possible.
> 
> Directly accessing parent variables means you have to know what is in the parent and what consistency requirements it has. Using an accessor means you only need to know the API and consistency is maintained for free.

That's an interesting point. I'll make the change, no problem.

Just because it's an interesting topic I love to discuss:
I disagree that Python, C++, Java (etc?) were wrong to add protected members, and I disagree that access by a subclass breaks encapsulation. You don't have to humor me with a reply, or perhaps we could have a nerdy coffee break sometime.

source/fab/steps/grab/fcm.py
bblay: Are you sure class variables work like that?

A class variable is evaluated _on declaration_, before any instantiation takes place, when the class itself is being evaluated. It's shared between all instances.

It's easy to check if it behaves the way you're suggesting:

```
from fab.steps.grab.fcm import FcmCheckout
from fab.steps.grab.svn import SvnCheckout

f1 = FcmCheckout('', '')
s1 = SvnCheckout('', '')

f2 = FcmCheckout('', '')
s2 = SvnCheckout('', '')

print(f1.command)
print(f2.command)
print(s1.command)
print(s2.command)
print()
print(f1.command)
print(s1.command)
print(f2.command)
print(s2.command)
```
```
fcm
fcm
svn
svn

fcm
svn
fcm
svn
```
We easily know exactly which command is being called by each class. There's no ambiguity, no race condition. This is how class variables work, unless I'm having a huge brain fart - entirely possible!

Calling a constructor won't change the value of a class variable, even by accident:
```
class C(object):
    att = 1
    def __init__(self):
        self.att = 2

c = C()
print(c.att)
print(c.__class__.att)
```
```
2
1
```

I'm happy to switch to dependency injection for the command, no problem at all.
But if the request is potentially based on a misunderstanding of Python, I think it's also useful for us to explore that, just to be sure.

source/fab/steps/grab/svn.py
bblay: That's an interesting point. I'll make the change, no problem, although this isn't a member variable, it's a method, so the solution would presumably be to make it public? That's a bit weird because we don't really need other classes to call this. Shall I go ahead with that anyway?

Just because it's an interesting topic I love to discuss:
I disagree that Python, C++, Java (etc?) were wrong to add protected variables, and I disagree that access by a subclass breaks encapsulation. You don't have to humor me with a reply, or perhaps we could have a nerdy coffee break sometime.

source/fab/steps/grab/svn.py
bblay: Ah, that's better and it works with fcm too, thank you.

source/fab/steps/grab/svn.py
MatthewHambley: In my zeal for eliminating the use of non-public members I overlooked that it was a method rather than a property. Still, is there any reason why it can't be public? Is there any reason why someone holding one of these objects shouldn't know if it is associated with a working copy?

MatthewHambley (changes requested):
Just a few final issues to think about.

source/fab/steps/grab/svn.py
bblay: I made it private to keep the interface small but there's no harm in making it public, will do.

MatthewHambley (approved):
Lets get this change on and get things moving.

PR #211, opened 2023-01-30
rsync dependency

bblay:
4 failing tests due to missing rsync.
unit tests and other system tests pass.


MatthewHambley (approved):
This looks like it adds rsync as a dependency.

Issue #219, opened 2023-02-08
False positives in "unused" list


MatthewHambley:
It seems like `program` program units are being incorrectly flagged as "unused". Is this because they are used by the linker, not by Fortran `use` statements?

PR #226, opened 2023-02-10
zero config

bblay:
First version of zero config command line tool.

Usage:
```
$ fab
$ fab <folder>
```

`automatically found the following programs to build: second, first`

Tested on:
 - some system test projects
    - FortranDependencies
    - CFortranInterop

So far, only tested using gfortran. That needs a linker flag, `-lgfortran`. We should try it with ifort, too. There might be a few linker flags we'll want to add to cover a wide variety of use cases.

todo:
 - [ ]  test it with other stuff
 - [ ]  test it with ifort


source/fab/cli.py
```diff
@@ -0,0 +1,83 @@
+# ##############################################################################
+#  (c) Crown copyright Met Office. All rights reserved.
+#  For further details please refer to the file COPYRIGHT
+#  which you should have received as part of this distribution
+# ##############################################################################
+from argparse import ArgumentParser
+from pathlib import Path
+
+from fab.steps.link import LinkExe
```
bblay, line 9: pycharm keeps doing this with auto imports 😠

source/fab/cli.py
```diff
@@ -0,0 +1,83 @@
...
+from fab.steps.root_inc_files import RootIncFiles
+
+
+def _generic_build_config(folder: Path, kwargs: Optional[Dict] = None) -> BuildConfig:
+    folder = folder.resolve()
+    kwargs = kwargs or {}
+
+    # Within the fab workspace, we'll create a project workspace.
+    # Ideally we'd just use folder.name, but to avoid clashes, we'll use the full absolute path.
+    label = '/'.join(folder.parts[1:])
```
MatthewHambley, line 31: Is there  a way to do this using only Path objects so as to avoid string handling? Is it important to do that if the resulting label is needed as a string anyway?

source/fab/cli.py
```diff
@@ -0,0 +1,83 @@
...
+
+    fpp, fpp_flags = get_fortran_preprocessor()
+    fc, fc_flags = get_fortran_compiler()
+
+    if fc == 'gfortran':
+        link_step = LinkExe(linker='gcc', flags=['-lgfortran'])
+    else:
+        raise NotImplementedError(f"Fab's zero config not yet configured for compiler: '{fc}'")
```
MatthewHambley, line 39: In my "mesh tools" script I made use of a dictionary to easily map compiler to argument list. Would that approach make sense here? e.g.

```python
fortran_linker = {
    'gfortran': ('gcc, ['-lgfortran']),
    'ifort': (..., [...])
}
...
try:
    link_step = LineExe(linker=fortran_link[fc][0], flags=fortran_link[fc][1])
except KeyError:
    raise NotImplementedError(f"Fab's zero configuration mode does not yet work with compiler '{fc}'"
```

source/fab/cli.py
```diff
@@ -0,0 +1,83 @@
...
+        steps=[
+            GrabFolder(folder),
+            FindSourceFiles(),
+
+            RootIncFiles(),  # JULES helper, get rid of this eventually
+
+            fortran_preprocessor(preprocessor=fpp, common_flags=fpp_flags),
+
+            CPragmaInjector(),
+            c_preprocessor(source=CollectionGetter(PRAGMAD_C)),
```
MatthewHambley, line 52: The zero configuration build mode should definitely look for C files. When it gains the ability to output a sample configuration script it should only include C (or Fortran) if they were used in the build.

source/fab/cli.py
```diff
@@ -0,0 +1,83 @@
...
+    return config
+
+
+def cli_fab():
+    """
+    Running Fab from the command line will attempt to build the project in the current or given folder.
+
+    """
+    arg_parser = ArgumentParser()
+    arg_parser.add_argument('folder', nargs='?', default='.', type=Path)
```
MatthewHambley, line 73: Setting the default to `Path()` will have the same effect. I don't know if it's any better, just thought I'd mention it.

source/fab/cli.py
```diff
@@ -0,0 +1,83 @@
...
+    arg_parser.add_argument('-v', '--version', action='store_true')
```
MatthewHambley, line 74: `argparse` provides a `version` action which means you don't have to handle this yourself.

source/fab/parse/fortran.py
```diff
@@ -185,7 +198,7 @@ def walk_nodes(self, fpath, file_hash, node_tree) -> AnalysedFortran:
                         analysed_fortran.add_symbol_dep(called_name.string)
 
                 elif obj_type == Program_Stmt:
-                    analysed_fortran.add_symbol_def(str(obj.get_name()))
+                    analysed_fortran.add_program_def(str(obj.get_name()))
```
MatthewHambley, line 201: This change might also fix the "unused symbols" check incorrectly listing program symbols. See #219.

source/fab/steps/analyse.py
```diff
@@ -167,6 +177,21 @@ def run(self, artefact_store: Dict, config):
...
+            # find fortran programs
+            sets_of_programs = [af.program_defs for af in by_type(analysed_files, AnalysedFortran)]
+            self.root_symbols = list(chain(*sets_of_programs))
+
+            # find c main()
+            c_with_main = list(filter(lambda c: 'main' in c.symbol_defs, by_type(analysed_files, AnalysedC)))
+            if c_with_main:
+                self.root_symbols.append('main')
+                if len(c_with_main) > 1:
+                    raise FabException("multiple c main() functions found")
```
MatthewHambley, line 191: It's okay for there to be multiple C programs defined by the source tree. If you don't want to handle that with this change then raise an issue or project card for it.

source/fab/steps/analyse.py
```diff
@@ -167,6 +177,21 @@ def run(self, artefact_store: Dict, config):
...
         self._add_manual_results(analysed_files)
 
+        # shall we search the results for fortran programs and a c function called main?
+        if self.find_programs:
+            # find fortran programs
+            sets_of_programs = [af.program_defs for af in by_type(analysed_files, AnalysedFortran)]
+            self.root_symbols = list(chain(*sets_of_programs))
+
+            # find c main()
+            c_with_main = list(filter(lambda c: 'main' in c.symbol_defs, by_type(analysed_files, AnalysedC)))
```
MatthewHambley, line 187: It may not be enough just to check for symbols called "main". I think we should be okay as class methods called "main" will get mangled but can you have a global variable called "main"? We don't necessarily need to handle these issues in this change but we should know what the possible failure modes are.

source/fab/steps/compile_fortran.py
```diff
@@ -346,6 +346,42 @@ def compile_file(self, analysed_file, flags, output_fpath):
             value={'time_taken': timer.taken, 'start': timer.start})
 
 
+def get_fortran_preprocessor():
+    fpp, fpp_flags = None, None
+
+    try:
+        fpp, fpp_flags = get_tool(os.getenv('FPP'))
+        logger.info(f'env defined fpp as "{fpp}"')
```
MatthewHambley, line 354: Shorthand is fine in debug messages but for "Info" level messages I think we prefer whole words and recognised contractions with apostrophes.

source/fab/steps/compile_fortran.py
```diff
@@ -346,6 +346,42 @@ def compile_file(self, analysed_file, flags, output_fpath):
...
+    except ValueError:
+        pass
+
+    if not fpp:
+        try:
+            run_command(['which', 'fpp'])
+            fpp, fpp_flags = 'fpp', ['-P']
```
MatthewHambley, line 361: A note about the purpose of "-P" and a link to https://github.com/stfc/fparser/issues/390

source/fab/steps/compile_fortran.py
```diff
@@ -346,6 +346,42 @@ def compile_file(self, analysed_file, flags, output_fpath):
...
+        try:
+            run_command(['which', 'cpp'])
+            fpp, fpp_flags = 'cpp', ['-traditional-cpp', '-P']
+            logger.info('detected cpp')
+        except RuntimeError:
+            # fpp not available
+            pass
+
+    if not fpp:
+        raise RuntimeError('no fortran preprocessor specified or discovered')
```
MatthewHambley, line 377: In the future we might consider a slicker implementation which takes a list of candidates and iterates until one succeeds.

source/fab/steps/compile_fortran.py
```diff
@@ -356,8 +392,38 @@ def get_fortran_compiler(compiler: Optional[str] = None):
...
+            pass
+
+    if not fortran_compiler:
+        try:
+            run_command(['ifort', '--help'])
+            fortran_compiler = 'ifort', []
+            logger.info('detected ifort')
+        except RuntimeError:
+            # gfortran not available
+            pass
```
MatthewHambley, line 421: The preference order for compilers is tricky. GFortran is faster to compile and more widely deployed but it bit rubbish. IFort is slower in compile and less common but produces much better executables. Some thought needs to be given this.

The fact that we go to the `FC` environment variable may be all the solution we need.

system-tests-new/zero_config/test_zero_config.py
```diff
@@ -0,0 +1,23 @@
...
+class TestZeroConfig(object):
+
+    def test_fortran_dependencies(self, tmp_path):
+        # test the sample project in the fortran dependencies system test
+        config = _generic_build_config(
+            Path(__file__).parent.parent / 'FortranDependencies',
+            kwargs={'fab_workspace': tmp_path, 'multiprocessing': False})
+        config.run()
+        assert (config.project_workspace / 'first.exe').exists()
+        assert (config.project_workspace / 'second.exe').exists()
```
MatthewHambley, line 15: Is it enough to test for the existence of an executable or do we need to run it as well?

tests/unit_tests/steps/test_compile_fortran.py
```diff
@@ -457,3 +458,71 @@ def test_vanilla(self):
...
+
+        assert fpp == 'foo_pp'
+        assert fpp_flags == ['--foo', '-P']
+
+    def test_empty_env_fpp(self):
+        def mock_run_command(command):
+            if 'fpp' not in command:
+                raise RuntimeError('foo')
+
+        with mock.patch.dict(os.environ, values={'FPP': ''}):
```
MatthewHambley, line 477: Does this remove the `FPP` variable from the environment  or simple set it to the empty string?

MatthewHambley (changes requested):
I'm excited to see movement on this mode of operation. This looks like a good starting point for further development.

source/fab/cli.py
bblay: We _could_ change `label` to be a path, as it's mainly used to create a subfolder in the fab workspace. I'm not sure we should bother doing that, so yes, it's probably not important, as you say.

source/fab/cli.py
bblay: Yes, that makes perfect sense. It's the natural progression as we add more compilers but we might as well do that right away.

source/fab/cli.py
bblay: ooh lovely, thanks.

source/fab/parse/fortran.py
bblay: Sadly not, as it calls `add_symbol_def` itself. Should still be an easy fix when we come to it though.

source/fab/steps/analyse.py
bblay: :+1: made [a ticket ](https://github.com/orgs/metomi/projects/1/views/1?pane=issue&itemId=21047392)

source/fab/steps/analyse.py
bblay: Agreed, although that feels like an extreme edge case.

system-tests-new/zero_config/test_zero_config.py
bblay: I think we should run all the exes we produce in our system tests, but I'm not sure it's a high priority yet.

tests/unit_tests/steps/test_compile_fortran.py
bblay: It sets it to an empty string.

bblay:
Just noticed a todo I'd put at the top, to test with ifort. Will do that.

source/fab/steps/compile_fortran.py
MatthewHambley: Is GitHub playing silly buggers? you've acknowledged this comment but I don't see a comment in the diff.

tests/unit_tests/steps/test_compile_fortran.py
```diff
@@ -457,3 +458,71 @@ def test_vanilla(self):
...
+
+        assert fpp == 'fpp'
+        assert fpp_flags == ['-P']
+
+    def test_empty_env_cpp(self):
+        def mock_run_command(command):
+            if 'cpp' not in command:
+                raise RuntimeError('foo')
+
+        with mock.patch.dict(os.environ, values={'FPP': ''}):
```
MatthewHambley, line 489: Another case of setting `FPP` to the empty string.

tests/unit_tests/steps/test_compile_fortran.py
```diff
@@ -457,3 +458,71 @@ def test_vanilla(self):
...
+
+        assert fc == 'foo_c'
+        assert fc_flags == ['--foo']
+
+    def test_empty_env_gfortran(self):
+        def mock_run_command(command):
+            if 'gfortran' not in command:
+                raise RuntimeError('foo')
+
+        with mock.patch.dict(os.environ, values={'FC': ''}):
```
MatthewHambley, line 511: And another.

tests/unit_tests/steps/test_compile_fortran.py
```diff
@@ -457,3 +458,71 @@ def test_vanilla(self):
...
+
+        assert fc == 'gfortran'
+        assert fc_flags == []
+
+    def test_empty_env_ifort(self):
+        def mock_run_command(command):
+            if 'ifort' not in command:
+                raise RuntimeError('foo')
+
+        with mock.patch.dict(os.environ, values={'FC': ''}):
```
MatthewHambley, line 523: More empty strings

MatthewHambley (changes requested):
Just a few more bits of tidying up.

source/fab/steps/compile_fortran.py
bblay: I think you'll find it is _I_ who is the silly bugger. Done now.

tests/unit_tests/steps/test_compile_fortran.py
bblay: :+1: thank you

tests/unit_tests/steps/test_compile_fortran.py
bblay: gosh

MatthewHambley (approved):
Time to bring this to trunk.

PR #228, opened 2023-02-15
verbose run configs

bblay:
We isolated these changes from the incremental psyclone pr to keep things smaller.
Some improvements to the example run configs, including a verbose argument which can be set in the rose config


source/fab/util.py
```diff
@@ -279,7 +279,7 @@ def common_arg_parser():
     """
     # consider adding preprocessor, linker, optimisation, two-stage
     arg_parser = ArgumentParser()
-    arg_parser.add_argument('--compiler', default=None)
+    arg_parser.add_argument('--verbose', action='store_true')
```
t00sa, line 282: Very much a question of CUI style: have you considered adding the common options to a separate group with add_argument_group?  Also, a help message for the verbose option would be good.

t00sa:
I'm generally happy with the PR.  I've added a comment about help messages on options and maybe using an argument group for default options, but I appreciate these are largely matters of taste and style.

source/fab/util.py
bblay: Hi Sam, thanks for the review!

We've added that, thanks for the suggestion. I'm not 100% sure about it, because the description of the group must currently be "_can_ be added to the BuildConfig" as it's up to the script dev to pass it through. But it's not causing any harm. Perhaps we could evolve this in the future to make it automatic. Here's the output:

```(sci-fab) ~/git/fab/run_configs/jules (run_config_changes) $ ./build_jules.py --help
usage: build_jules.py [-h] [--verbose] [--two-stage] [--revision REVISION]

optional arguments:
  -h, --help           show this help message and exit
  --revision REVISION

common arguments:
  Common arguments which can be passed to the BuildConfig.

  --verbose            DEBUG level logging
  --two-stage          Compile .mod files first in a separate pass.
                       Theoretically faster in some projects.
```

t00sa: Thank you - that looks good.  I like the idea of making things automatic: if the users get the same look-and-feel from the command line tools, it makes everyone's life easier and happier.

t00sa (approved):
I'm happy with these changes and I think they're OK to merge.

bblay:
Thanks Sam!

PR #233, opened 2023-02-27
pskalite working concept

bblay:
pskalite working concept


source/fab/steps/psyclone.py
```diff
@@ -57,6 +57,7 @@ class MpPayload:
     """
     analysed_x90: Dict[Path, AnalysedX90]
     all_kernel_hashes: Dict[str, int]
+    override_files: List[str]  # filenames (not paths) of hand crafted overrides
```
MatthewHambley, line 60: The Path object can hold relative paths. Is that what is needed here or do you need just the name of the file?

source/fab/steps/psyclone.py
```diff
@@ -79,7 +80,24 @@ class Psyclone(Step):
...
+            Folder containing kernel files. Must be part of the analysed source code.
+        :param transformation_script:
+            The Python transformation script.
+        :param cli_args:
+            Passed through to the psyclone cli tool.
+        :param source_getter:
+            Optional override for getting input files from the artefact store.
+        :param overrides_folder:
+            Optional folder containing hand-crafted override files.
+            Must be part of the analysed source code.
```
MatthewHambley, line 98: Is there a potential problem with having two files which define the same symbols? Can you think of any scenario when there might be two copies of the symbol floating around.

If the source has to have been analysed before override happens doesn't that mean there's a copy of the symbol in the override file and one in the generated (and then overridden) output?

MatthewHambley (changes requested):
Just a few questions really.

source/fab/steps/psyclone.py
bblay: We just need the name of the file.

Currently, we find an override by comparing just the file names, then construct the override path as `override_folder / filename`, which assumes the override folder is flat.

source/fab/steps/psyclone.py
bblay: > If the source has to have been analysed before override happens

It's the other way round, we have to run psyclone before we run the analysis step because the psyclone step produces new Fortran. We'll make this comment clearer.

MatthewHambley (approved):
There is a shortage of system testing for this new feature but that is part of a wider discussion about system and acceptance testing.

PR #237, opened 2023-03-03
tidy test folders

bblay:
We now have a `tests` folder containing all current unit and system tests, plus a `tests-old` folder containing as-yet unattended older unit tests, system tests and test cases`. All test files are in these two folders now.


mo-lucy-gordon (dismissed):
Changes look good to me

bblay:
@mo-lucy-gordon, this needs re-review/approval after merging in the lastest master because a new test had been added since this branch was created, and had to be moved in with the others.

mo-lucy-gordon (commented):
@bblay Like you mentioned, I think it would be better to have consistency with the naming of unit_tests and system-tests.

mo-lucy-gordon (approved):

PR #243, opened 2023-03-28
preprocess copys little f90s across

bblay:
The fortran preprocessor copies little f90s across to the build output.

Motivation: In the latest lfric, some kernels are little f90s. We can't tell psyclone to look in both the source and build output folders for kernels because it detects duplicates. This change ensures all preprocessed fortran is in one place.

todo:
 - [ ] tests

### DIFF
--- a/source/fab/steps/preprocess.py
+++ b/source/fab/steps/preprocess.py
@@ -9,15 +9,16 @@ Fortran and C Preprocessing.
 """
 import logging
 import os
+import shutil
 from dataclasses import dataclass
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import Collection, Iterable, List, Optional, Tuple
 
 from fab.build_config import BuildConfig, FlagsConfig
 from fab.constants import PRAGMAD_C
 from fab.metrics import send_metric
 
-from fab.util import log_or_dot_finish, input_to_output_fpath, log_or_dot, Timer, by_type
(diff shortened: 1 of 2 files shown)


source/fab/steps/preprocess.py
```diff
@@ -36,7 +37,7 @@ class MpCommonArgs(object):
 
 
 def pre_processor(config: BuildConfig, preprocessor: str,
-                  source_getter: ArtefactsGetter, output_collection, output_suffix,
+                  files: Collection[Path], output_collection, output_suffix,
```
MatthewHambley, line 40: Why the change from "getter" to list of files? Do we not want to get these artefacts any more?

MatthewHambley (changes requested):
Check documentation pages to make sure there are no changes needed there.

source/fab/steps/preprocess.py
bblay: Because, in the fortran preprocessor which wraps this, we want to get the list of files for the getter once, then filter it for big and little f90. This keeps the capability for the user to specify the getter themselves, or for us to provide a default, and for whichever it is to be called once.

MatthewHambley (changes requested):
I just realised, there are no tests of this change!

MatthewHambley (approved):
That's enough testing for us to move on to a release.

PR #244, opened 2023-03-29
Updated step func docs

wilbertcs:
Documentation updated with 'step_func' changes.


docs/source/advanced_config.rst
```diff
@@ -166,21 +164,19 @@ The C preprocessor looks for the output of this step by default.
...
-
+    ...
+    c_pragma_injector(config)
+    preprocess_c(config)
+    ...
 
 Custom Steps
 ============
-If you need a custom build step, you can create a subclass of the :class:`~fab.steps.Step` class.
+If you need a custom build step, you can create a function with the @step_timer decorator.
```
MatthewHambley, line 176: For later, a better name which doesn't mention timers? It doesn't seem like a useful concept to include and is potentially confusin.

MatthewHambley (approved):
This is actually a fairly simple change of converting references to classes into references to functions.

PR #252, opened 2023-04-03
Polish the documentation

MatthewHambley:
Early indications are that the installation documentation is confusing people. I've tried to tighten it up a bit and made a light pass on everything else while I was at it.


docs/source/metoffice.rst
```diff
@@ -0,0 +1,165 @@
...
+Fortran Only
+------------
+
+If you don't need to build C, you can use this the modules in the "extra
+software" collection:
+
+.. code-block:: console
+
+    $ module use $EXTRA_SOFTWARE
+    $ module load python/3.7.0 support/fparser/0.16.0/python/3.7.0 fab/0.11.0/python/3.7.0
```
lianneparkhouse, line 25: This `module load` command (`module load python/3.7.0 support/fparser/0.16.0/python/3.7.0 fab/0.11.0/python/3.7.0`) does not seem to work for me. I receive the following error(s): 

ModuleCmd_Load.c(213):ERROR:105: Unable to locate a modulefile for 'python/3.7.0'
ModuleCmd_Load.c(213):ERROR:105: Unable to locate a modulefile for 'support/fparser/0.16.0/python/3.7.0'
ModuleCmd_Load.c(213):ERROR:105: Unable to locate a modulefile for 'fab/0.11.0/python/3.7.0'

May I be missing something here? I managed to successfully execute `module use $EXTRA_SOFTWARE` (with the correct path, provided by Giorgia).

docs/source/metoffice.rst
```diff
@@ -0,0 +1,165 @@
...
+    $ pip install --upgrade pip
+    $ pip install Jinja2
+    $ pip install six
+    $ pip install fparser
+    $ pip install pyparsing
+    $ pip install sympy
+    $ pip install jsonschema==3.0.2
+    $ pip install configparser
+
+Finally install the Fab in your working copy. This is done in "editable" mode
```
lianneparkhouse, line 107: Very minor comment, but should "the" be removed from here? So:
```suggestion
Finally, install Fab in your working copy. This is done in "editable" mode
```

docs/source/environment.rst
```diff
@@ -2,150 +2,131 @@
...
 
-Then :ref:`install fab<Install>`.
+Somebody Elses Problem
+----------------------
 
-We do this in a before any module commands.
-Then we set up our environment *in a new terminal* as follows.
+Some of the prerequisites may require system administrator intervention. If they
+are feeling benevolent your administrator may even be prevailed upon to install
+the whole lot.
```
lianneparkhouse, line 54: I like the title of this section (😄), does this mean that I can contact an admin to create the environment for me, or would they install all of Fab for me? Would this be a ServiceNow request for MO users?

docs/source/metoffice.rst
```diff
@@ -0,0 +1,165 @@
...
+Using Fab at the Met Office
+===========================
+
+For users working on Met Office desktop systems the following instructions may
+prove useful.
+
+Using Fab
+~~~~~~~~~
+
+If you just want to use Fab to build some source then there are a number of
```
lianneparkhouse, line 12: I wasn't too sure about this sentence – might this be a typo?

```suggestion
If you just want to use Fab to build from source then there are a number of
```

GiorgiaLine (changes requested):
Changes requested by second reviewer, passing back for further work.

docs/source/metoffice.rst
MatthewHambley: Not sure why Python didn't load, it works for me. fParser had the wrong version number,  not fixed.

I notice that Fab is still the beta version so I've updated it to 1.0.0 but I'm still waiting for that module to be reviewed so it doesn't exist yet. The beta module listed here does exist so I'm not sure why it fails.

docs/source/metoffice.rst
MatthewHambley: Done

docs/source/environment.rst
MatthewHambley: This is general documentation for all users. In the case of MO, no, there is no-one else who will do this for you.

docs/source/metoffice.rst
MatthewHambley: Tried an alternative.

GiorgiaLine (approved):
Thanks for those changes, looks good. Happy to approve for merge.

docs/source/metoffice.rst
```diff
@@ -0,0 +1,204 @@
...
+
+If you don't need to build C, you can use this the modules in the "extra
+software" collection:
+
+.. code-block:: console
+
+    $ module use $EXTRA_SOFTWARE_MODULE_DIR
+    $ module load python/3.7.0 support/fparser/0.0.16/python/3.7.0 fab/1.0.0/python/3.7.0
+
+The Science I.T. group can provide the `EXTRA_SOFTWARE` path.
```
lianneparkhouse, line 27: Now that I can successfully load the correct modules, is the next step then to simply install Fab? If so, I think a quick "Then [install fab](https://metomi.github.io/fab/install.html#install)" at the end of the **Fortran only** section might be useful here for new users, as I wasn't too sure which step to follow next (if I don't need to build C).

lianneparkhouse (changes requested):
Thanks for your replies, @MatthewHambley. I only have one final query.

docs/source/metoffice.rst
MatthewHambley: I've added a sentence to make explicit what loading the module achieves.

lianneparkhouse (approved):
That all makes sense now, thank you @MatthewHambley. I think these improved instructions are clear and all works for me. **Approved**

PR #254, opened 2023-04-18
Tidy CLI parsing and zero config

wilbertcs:
Cleaned up argument parsing for zero config user scripts and testing.

### DIFF
--- a/source/fab/cli.py
+++ b/source/fab/cli.py
@@ -22,7 +22,7 @@ from fab.steps.preprocess import preprocess_c, preprocess_fortran
 from fab.util import common_arg_parser
 
 
-def _generic_build_config(folder: str = '.', kwargs=None) -> BuildConfig:
+def _generic_build_config(folder: Path, kwargs=None) -> BuildConfig:
     project_label = 'zero_config_build'
     if kwargs:
         project_label = kwargs.pop('project_label', 'zero_config_build') or project_label
@@ -69,7 +69,7 @@ def calc_linker_flags():
     return linker, linker_flags
 
 
-def cli_fab(folder: str = '.', kwargs: Optional[Dict] = None):
+def cli_fab(folder: Path, kwargs: Optional[Dict] = None):
     """
     Running Fab from the command line will attempt to build the project in the current or given folder.
     The following params are used for testing. When run normally any parameters will be caught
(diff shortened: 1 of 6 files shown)


source/fab/cli.py
```diff
@@ -12,28 +12,25 @@
...
 from fab.steps.find_source_files import find_source_files
 from fab.steps.grab.folder import grab_folder
 from fab.steps.preprocess import preprocess_c, preprocess_fortran
+from fab.util import common_arg_parser
 
 
-def _generic_build_config(folder: Path, kwargs: Optional[Dict] = None) -> BuildConfig:
-    folder = folder.resolve()
-    kwargs = kwargs or {}
+def _generic_build_config(folder: str = '.', kwargs=None) -> BuildConfig:
```
MatthewHambley, line 25: What was the reason for falling back to path-as-string? If we can keep them as Path objects we should. The default constructor for Path gives the currently selected directory (e.g. `Path()`) if that was the problem.

source/fab/cli.py
```diff
@@ -72,15 +69,24 @@ def calc_linker_flags():
...
+        parameters  ( Testing Only )
 
     """
-    # todo: use common_arg_parser()?
-    arg_parser = ArgumentParser()
-    arg_parser.add_argument('folder', nargs='?', default='.', type=Path)
-    arg_parser.add_argument('--version', action='version', version=f'%(prog)s {fab.__version__}')
-    args = arg_parser.parse_args()
+    kwargs = kwargs or {}
+    if Path(sys.argv[0]).parts[-1] == 'fab':
```
MatthewHambley, line 85: This seems to be checking that the executable is called as "fab". Is this correct and under what circumstances would it not be called that?

MatthewHambley (changes requested):
Just a few questions really.

source/fab/cli.py
wilbertcs: 'fab' could , and is called by other things like 'pytest' so therefore the cli arguments may not apply to 'fab' which causes the argparser to fail with an invalid argument message.

source/fab/cli.py
wilbertcs: This is only used to create the 'project_label' when using 'zero config' fab, so it simplifies the process.

MatthewHambley (changes requested):
That seems reasonable, please make the change you propose and add a comment to explain the name check.

source/fab/cli.py
wilbertcs: My mistake it should have been a Path object

MatthewHambley (approved):
Looks like it's ready to go now.

PR #256, opened 2023-04-24
Remove '.ext' ext from built executables

wilbertcs:


GiorgiaLine (approved):
Does what it says on the tin. Approved for merge.

PR #267, opened 2023-08-16
Correct errors in build configuration examples

t00sa:
The examples in the build configuration page contain a number of typos and errors, mostly related to imports and design changes prior to fab 1.0.  This updates the page to fix the bugs.


MatthewHambley (approved):
Not much to complain about in these changes, they all look fine.

Of course none of the C.I. tests passed because something seems to have broken there. This is just a change to documentation so I will force the issue.

PR #268, opened 2023-08-17
Remove deprecated type comparisons

MatthewHambley:
The most recent version of flake8 running on Python 3.10 complains about comparisons in the form `type(thing) == ThingType` and requests they be changed to `isinstance()` calls.


GiorgiaLine (approved):
Looks good

PR #279, opened 2024-03-02
278 absolute paths in compilation

### LINKED ISSUE #278: Use absolute path when compiling files
A very minor improvement:

ATM a relative path is used for the source file when calling the compiler (i.e. only the file name). This works because the subprocess is changing into the right directory when executing the compiler.

But this makes it harder to verify if the right files are compiled (e.g. I spent some time debugging based on the assumption it would compile a different version of the file, since I wasn't sure of the implicit 'chdir' done), and also does not allow you to simply copy&paste the compilation line (e.g. for testing).

I have a patch for this ready (but apparently can neither assign this to me, nor set a label :) ).

hiker:
A tiny tiny change to make the Fortran compilation use absolute paths, the same way c compilation does. Besides consistency, it's also easier to debug if you can copy&paste the compilation lines :)

Fixes #278.

### DIFF
--- a/source/fab/steps/compile_fortran.py
+++ b/source/fab/steps/compile_fortran.py
@@ -363,7 +363,7 @@ def compile_file(analysed_file, flags, output_fpath, mp_common_args):
     command.append(str(analysed_file.fpath))
     command.extend(['-o', str(output_fpath)])
 
-    run_command(command, cwd=analysed_file.fpath.parent)
+    run_command(command)
 
 
 # todo: move this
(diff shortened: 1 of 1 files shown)


hiker:
Dang, as far as I can tell this does not work: gfortran adds the (now absolute) path to the .o file, which means that two identical compiled files in different workspaces now have a different hash (and `test_repeatable_fmod_hashes ` in `prebuild/test_prebuild.py` now fails). 
The only other idea I have for now is to modify the debug log message to indicate if the directory is changed.

hiker:
I am closing this PR for now, since it causes a test failure in the computed hashes. This is strongly related to #289  - if this issue can be fixed the way suggested in the ticket, then the current PR would work. If #289 cannot be fixed, we need a different approach here, e.g. instead of using absolute paths logging the cwd when executing Fortran, so you are aware of the cwd when debugging.

PR #282, opened 2024-03-02
Minor documentation update

### LINKED ISSUE #277: Minor Documentation Update
Hi,
while reading your documentation, I noticed a few very minor errors. I'll collect them all here (please let me know if there is a better location):

### 1. https://metomi.github.io/fab/writing_config.html
Under 'further reading' there is
> More advanced configuration topics are discussed in Advanced Configuration.

It would be great if this could become a link

### 2. https://metomi.github.io/fab/advanced_config.html
This page contains as the second code example:
```
from my_grab import my_grab_config

if __name__ == '__main__':
    with BuildConfig(project_label='<project_label>') as state:
        grab_folder(state, src=grab_config.source_root),
```
I believe the last line should use **my**_grab_config
``` 
        grab_folder(state, src=my_grab_config.source_root),
```

### https://metomi.github.io/fab/advanced_config.html#housekeeping
Not sure if I get that. FAB will automatically remove old files. In order to prevent that, you need to run your own cleanup step? I think an example would be great here. I would expect that if I run a 'cleanup' script, it would remove things earlier (or more things), since FAB would still do its own cleanup anyway. So, FAB detects that you have your own cleanup step, and then does not run its own cleanup? Is that correct?
[edit]: OK, I found it explicitly mentioned elsewhere (can't remember where). I would suggest to update the house keeping paragraph to include:

> You can add a [cleanup_prebuilds()](https://metomi.github.io/fab/apidoc/fab.steps.cleanup_prebuilds.html#fab.steps.cleanup_prebuilds.cleanup_prebuilds) step, where you can explicitly control how long to keep prebuild files. This may be useful, for example, if you often switch between two versions of your code and want to keep the prebuild speed benefits when building both. If you do not add your own cleanup_prebuild step, Fab will automatically run a default step which will remove all prebuild files that are not part of the current build.

### https://metomi.github.io/fab/apidoc/fab.steps.compile_fortran.html
This contains:
> source ([Optional](https://docs.python.org/3/library/typing.html#typing.Optional)[[ArtefactsGetter](https://metomi.github.io/fab/apidoc/fab.artefacts.html#fab.artefacts.ArtefactsGetter)]) – (default: None) An [ArtefactsGetter](https://metomi.github.io/fab/apidoc/fab.artefacts.html#fab.artefacts.ArtefactsGetter) which give us our c files to process.

Notice the **c** files mentioned :)

hiker:
Minor documentation fixes for #277. It additionally fixes all other broken links reported (I had to guess a bit where the 'advanced c code' link was supposed to go - please let me know if not to the pragma injector section.).

Fixes #277.

### DIFF
--- a/docs/source/advanced_config.rst
+++ b/docs/source/advanced_config.rst
@@ -82,18 +82,20 @@ import your grab configuration to find out where it put the source.
 
     if __name__ == '__main__':
         with BuildConfig(project_label='<project_label>') as state:
-            grab_folder(state, src=grab_config.source_root),
+            grab_folder(state, src=my_grab_config.source_root),
 
 
 Housekeeping
 ============
 
-Fab will remove old files from the prebuilds folder. It will remove all prebuild files that are not part of the current build by default.
-
-If you add a :func:`~fab.steps.cleanup_prebuilds.cleanup_prebuilds` step, you
-can keep prebuild files for longer. This may be useful, for example, if you
-often switch between two versions of your code and want to keep the prebuild
-speed benefits when building both.
+You can add a :func:`~fab.steps.cleanup_prebuilds.cleanup_prebuilds`
+step, where you can explicitly control how long to keep prebuild files.
+This may be useful, for example, if you often switch between two versions
+of your code and want to keep the prebuild speed benefits when building
+both. If you do not add your own cleanup_prebuild step, Fab will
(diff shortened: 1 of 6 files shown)


hiker:
Oops - closing this for now, it was based on the wrong branch and has additional changes :(

PR #291, opened 2024-03-14
Fix test error

jasonjunweilyu:
This PR fix all the test errors and has passed the CI pipeline. The detailed error messages are listed in the comment section of each commit and also shown in Issue hiker#4. Duplicated test names under subdirectories `system_tests` and `unit_tests` are renamed. Now with pytest, the tests can be run with `pytest tests` for all the tests and `pytest tests/unit_tests` or `pytest tests/system_tests` for a specific test group.

This is related to the original pull request https://github.com/hiker/fab/pull/9 on hiker/fab.


tests/system_tests/psyclone/test_psyclone_system_test.py
```diff
@@ -160,17 +161,17 @@ def test_run(self, config):
...
-            config.prebuild_folder / 'algorithm_mod_psy.5088673431.f90',  # prebuild
+            # The kernel hash differs between fpp and cpp, so just use wildcards.
+            config.prebuild_folder / 'algorithm_mod.*.an',  # x90 analysis result
+            config.prebuild_folder / 'my_kernel_mod.*.an',  # kernel analysis results
+            config.prebuild_folder / 'algorithm_mod.*.f90',  # prebuild
+            config.prebuild_folder / 'algorithm_mod_psy.*.f90',  # prebuild
         ]
 
-        assert all(not f.exists() for f in expect_files)
+        assert all((glob.glob(str(f)) == []) for f in expect_files)
```
MatthewHambley, line 171: There is a "glob" method on `Path` objects. Might be neater than casting to string?

tests/system_tests/psyclone/test_psyclone_system_test.py
```diff
@@ -160,17 +161,17 @@ def test_run(self, config):
...
         with config:
             self.steps(config)
-        assert all(f.exists() for f in expect_files)
+        assert all((glob.glob(str(f)) != []) for f in expect_files)
```
MatthewHambley, line 174: See previous comment.

MatthewHambley (changes requested):
There is a general issue regarding the value of the current system tests which this change does not tackle. It does tidy things up a bit so I don't see any reason not to go ahead with it anyway.

tests/system_tests/psyclone/test_psyclone_system_test.py
hiker: I actually discussed this with Jason, the main issue we had was that while `Path` has a `glob` method, it doesn't work the way I expected it to do: 
```
p = Path("/tmp/*.txt")
p.glob()   # error :(
```
because `glob` requires a pattern, it does not automatically take the pattern from the path.

So assuming that you are ok with it I will change this as:
```
p.parent.glob(p.name) ...
```
This seems to work

hiker: I had to use a `list(...)`, since a generator (which is returned from Path's glob method) cannot easily be tested to see if it is empty.

tests/system_tests/psyclone/test_psyclone_system_test.py
hiker: Ditto, used Path's glob method, and a list (since otherwise the test will compare if an generator is != [] ... which is the case even if the iterator return nothing).

hiker:
Ready for next review.

tests/system_tests/psyclone/test_psyclone_system_test.py
```diff
@@ -160,17 +160,19 @@ def test_run(self, config):
...
+            config.prebuild_folder / 'algorithm_mod.*.an',  # x90 analysis result
+            config.prebuild_folder / 'my_kernel_mod.*.an',  # kernel analysis results
+            config.prebuild_folder / 'algorithm_mod.*.f90',  # prebuild
+            config.prebuild_folder / 'algorithm_mod_psy.*.f90',  # prebuild
         ]
 
-        assert all(not f.exists() for f in expect_files)
+        # Glob returns a generator, which can't simply be tested if it's empty.
+        # So use a list instead:
+        assert all(list(f.parent.glob(f.name)) == [] for f in expect_files)
```
MatthewHambley, line 172: An alternative solution would be to split the expected results by directory.

```
expected_prebuild_files = ['algorithm_mod.*.an', 'my_kernel_mod.*.an', ...]
assert all(list(config.prebuild_folder.glob(pattern)) == [] for pattern in expected_prebuild_files

expected_build_files = ['algorithm/algorithm_mod.f90', ...
```

This might be neater. See what you think.

MatthewHambley (changes requested):
A potential alternative which might be neater.

Please use the "re-request review" icon next to my username in the reviewers list when you're ready for me to have another look.

tests/system_tests/psyclone/test_psyclone_system_test.py
jasonjunweilyu: Hi Matthew,

I am back from holiday. Thanks for the suggestion. I have implemented this in the newest commit. Now it is ready for review.

Best regards,
Jason

MatthewHambley (approved):
Happy for this change to proceed.

PR #298, opened 2024-03-28
11 rsync links

### LINKED ISSUE #290: rsync should copy links
The `rsync` command used in FAB does not copy links.

Recently, LFRic has started to use links for the PSyclone optimisation scripts (as the easiest way of making sure the various scripts from all partner sights stay up-to-date - while one might not agree with that, for now this is the way it is). If you now use `grab_folder` to copy the optimisation scripts, these links are missing, and the optimisation script cannot be found.

Simple fix: add `--links` to the rsync command.

hiker:
This adds a flag to rsync to include links, which is required for newer LFRic versions which use links for identical optimisation scripts. Fixes #290.


hiker:
@MatthewHambley sorry, I can't neither request a review, nor change its label or anything :) This PR is ready to be reviewed

MatthewHambley (approved):
Does what it says on the tin.

PR #323, opened 2024-07-10
Restyle repository to 80 columns.

MatthewHambley:
Simply bumping up the acceptable line width doesn't seem like a good way to tackle the problem of overly nested structures.

This change removes the special settings and restyles everything to 80 columns with exceptions where doing so would tie us in knots.

This change does not attempt to tackle any of the problems such as the powerful odour coming from all the mockery in unit tests.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-max-line-length = 120
-exclude = .git,.github,.gitignore,.idea,_cache,.run,.vscode,__pycache__,docs,venv,tests-old
--- a/Documentation/source/conf.py
+++ b/Documentation/source/conf.py
@@ -88,7 +88,8 @@ html_theme_options = {
 html_static_path = ['_static']
 
 
-# Don't sort module contents alphabetically, we want to order them in a helpful way for the user to read.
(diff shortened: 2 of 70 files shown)


hiker:
While I totally agree with the 80 column column width, can we wait with this? We are at least 4 PRs ahead of trunk (and I have been reformatting things to 80 columns :) ), so please let us check what kind of conflicts this creates for us. I assume this is done with black (or a similar tool??), so we could do this on our end as the PR after if we have too many conflicts?

MatthewHambley:
I have no problem with pulling this change if it messes with what you are doing.

hiker:
It causes over 20 conflicts for us, not sure how much effort it would be to fix them :(  Did you just run `black` (or any other tool) on the sources? If we can re-run whatever you did, I would prefer that. But if you have done it manually, then we might have to bite the bullet and resolve the conflicts :(
FWIW, we are currently trying to change our fork to point back to this repo, atm we are 'disconnected', and can't even easily submit PRs anymore

MatthewHambley:
This change is being made piecemeal as other changes are made.

PR #325, opened 2024-07-17
Improve artefact workflow

hiker:
This PR introduces new artefact categories defined as enums. The code (and its typing) supports strings and this enum as keys, meaning any user script can artefacts as they want (and be certain not to interfere with the ones used in Fab).

I then used three new categories for fortan/c/psyclone build files, and updates all steps to keep these categories up-to-date. E.g. when pre-processing a Fortran file, the .F90 would be removed from the fortran-build-files, and the .f90 will be added etc. PSyclone will add the newly created files to the fortran-build-files. This means that e.g. the Fortran analysis step only need to look at two artefacts (C and Fortran), instead of the current one:
```
DEFAULT_SOURCE_GETTER = CollectionConcat([
    SuffixFilter('all_source', '.f90'),
    'preprocessed_c',
    'preprocessed_fortran',

    # todo: this is lfric stuff so might be better placed elsewhere
    SuffixFilter('psyclone_output', '.f90'),
    'preprocessed_psyclone',  # todo: this is no longer a collection, remove
    'configurator_output',
])
```
New version is now:
```
DEFAULT_SOURCE_GETTER = CollectionConcat([
    ArtefactSet.FORTRAN_BUILD_FILES,
    ArtefactSet.C_BUILD_FILES,
])
```

This also makes it MUCH easier to add a step (since there is no need to know the internals of Fab to figure out in which artefacts to find the required files). It also fixes (if I am not mistaken) the old approach where .f90 files are copied from src to build during preprocessing, BUT they are never added to an artefact, so the compilation will still be based on files in source.


hiker:
This is now ready. The PR looks so big because the change to the artefact names affected many files, plus I started to fix up files with re to coding standard like line lengths etc.

Documentation/source/writing_config.rst
```diff
@@ -269,7 +268,42 @@ then creates the executable.
...
+(or change them all):
+
+..
+  My apologies for the LONG lines, they were the only way I could find
+  to have properly indented paragraphs :(
+
+1. :func:`~fab.steps.find_source_files.find_source_files` will add all source files it finds to ``ALL_SOURCE`` (by default, can be overwritten by the user). Any ``.F90`` and ``.f90`` file will also be added to ``FORTRAN_BUILD_FILES``, any ``.c`` file to ``C_BUILD_FILES``, and any ``.x90`` or ``.X90`` file to ``X90_BUILD_FILES``. It can be called several times if files from different root directories need to be added, and it will automatically update the ``*_BUILD_FILES`` sets.
+2. Any user script that creates new files can add files to ``ALL_SOURCE`` if required, but also to the corresponding ``*_BUILD_FILES``. This will happen automatically if :func:`~fab.steps.find_source_files.find_source_files` is called to add these newly created files.
+3. If :func:`~fab.steps.c_pragma_injector.c_pragma_injector` is being called, it will handle all files in ``C_BUILD_FILES``, and will replace all the original C files with the newly created ones. For backward compatibility it will also store the new objects in the ``PRAGMAD_C`` set.
```
MatthewHambley, line 298: In case it hasn't been mentioned, given that we are using `libclang` to parse C there may be a more elegant way of not disappearing down a rabbit hole with system inclusions.

Documentation/source/writing_config.rst
```diff
@@ -269,7 +268,42 @@ then creates the executable.
...
+6. If :func:`~fab.steps.psyclone.preprocess_x90` is called, it will similarly preprocess all ``.X90`` files in ``X90_BUILD_FILES``, creating the output files in the build folder, and replacing the files in ``X90_BUILD_FILES``.
+7. If :func:`~fab.steps.psyclone.psyclone` is called, it will process all files in ``X90_BUILD_FILES`` and add any newly created file to ``FORTRAN_BUILD_FILES``, and removing them from ``X90_BUILD_FILES``.
+8. The :func:`~fab.steps.analyse.analyse` step analyses all files in ``FORTRAN_BUILD_FILES`` and ``C_BUILD_FILES``, and add all dependencies to ``BUILD_TREES``.
+9. The :func:`~fab.steps.compile_c.compile_c` and :func:`~fab.steps.compile_fortran.compile_fortran` steps will compile all files from ``C_BUILD_FILES`` and ``FORTRAN_BUILD_FILES``, and add them to ``OBJECT_FILES``.
+10. If :func:`~fab.steps.archive_objects.archive_objects` is called, it will create libraries based on ``OBJECT_FILES``, adding the libraries to ``OBJECT_ARCHIVES``.
+11. If :func:`~fab.steps.link.link_exe` is called, it will either use ``OBJECT_ARCHIVES``, or if this is empty, use ``OBJECT_FILES``, create the binaries, and add them to ``EXECUTABLES``.
```
MatthewHambley, line 306: Overall I appreciate the additional documentation on exactly what is happening. This was missing before.

run_configs/lfric/lfric_common.py
```diff
@@ -21,79 +23,113 @@ def check_available(self):
         return True
 
 
-# todo: is this part of psyclone? if so, put  it in the psyclone step module?
```
MatthewHambley, line 24: Just to be clear (and I think you understand this already) "Configurator" is its own tool unconnected to PSyclone.

run_configs/lfric/lfric_common.py
```diff
@@ -21,79 +23,113 @@ def check_available(self):
...
+    find_source_files(config, source_root=config_dir)
 
 
+# ============================================================================
 @step
 def fparser_workaround_stop_concatenation(config):
     """
-    fparser can't handle string concat in a stop statement. This step is a workaround.
+    fparser can't handle string concat in a stop statement. This step is
+    a workaround.
```
MatthewHambley, line 88: I wonder if this is still needed? If you are feeling diligent you might find the fParser issue number and include it in the comment.

run_configs/lfric/lfric_common.py
```diff
@@ -21,79 +23,113 @@ def check_available(self):
...
     bad = "_config: '// &\n        'Unable to close temporary file'"
     good = "_config: Unable to close temporary file'"
 
-    open(feign_config_mod_fpath, 'wt').write(
+    open(feign_path, 'wt').write(
         open(broken_version, 'rt').read().replace(bad, good))
+
+
+# ============================================================================
+def get_transformation_script(fpath, config):
```
MatthewHambley, line 114: Type hinting? Would remove the need for `rtype` in the docstring.

source/fab/parse/fortran.py
```diff
@@ -301,7 +301,11 @@ def _process_comment(self, analysed_file, obj):
             # TODO #13: once fparser supports reading the sentinels,
```
MatthewHambley, line 301: An fParser issue number in the comment would help tracking this.

source/fab/steps/archive_objects.py
```diff
@@ -49,46 +51,54 @@ def archive_objects(config: BuildConfig,
...
     **Creating a Static or Shared Library:**
 
-    When building a library there is expected to be a single build target with a `None` name.
-    This typically happens when configuring the :class:`~fab.steps.analyser.Analyser` step *without* a root symbol.
-    We can assume the list of object files is the entire project source, compiled.
+    When building a library there is expected to be a single build target
+    with a `None` name. This typically happens when configuring the
+    :class:`~fab.steps.analyser.Analyser` step *without* a root symbol.
+    We can assume the list of object files is the entire project source,
+    compiled.
```
MatthewHambley, line 66: This is true for the moment but I have a plan to use source directives to mark up the API which would allow "zero configuration" mode to produce multiple libraries. I suppose it's "implied configuration mode" at that point.

source/fab/steps/archive_objects.py
```diff
@@ -49,46 +51,54 @@ def archive_objects(config: BuildConfig,
...
     **Creating Executables:**
 
-    When creating executables, there is expected to be one or more build targets, each with a name.
-    This typically happens when configuring the :class:`~fab.steps.analyser.Analyser` step *with* a root symbol(s).
-    We can assume each list of object files is sufficient to build each *<root_symbol>.exe*.
+    When creating executables, there is expected to be one or more build
+    targets, each with a name. This typically happens when configuring the
+    :class:`~fab.steps.analyser.Analyser` step *with* a root symbol(s).
+    We can assume each list of object files is sufficient to build each
+    *<root_symbol>.exe*.
```
MatthewHambley, line 76: I don't know how best to deal with this in the documentation but `.exe` is a uniquely Windows concept.

tests/unit_tests/test_artefacts.py
```diff
@@ -1,10 +1,96 @@
...
+                           CollectionConcat, CollectionGetter,
+                           FilterBuildTrees, SuffixFilter)
+
+
+def test_artefact_store() -> None:
+    '''Tests the ArtefactStore class.'''
+    artefact_store = ArtefactStore()
+    assert len(artefact_store) == len(ArtefactSet)
+    assert isinstance(artefact_store, dict)
+    assert ArtefactSet.CURRENT_PREBUILDS in artefact_store
```
MatthewHambley, line 19: Why is `CURRENT_PREBUILDS` singled out for special testing?

MatthewHambley (changes requested):
This review turned into a real stream of consciousness so probably best to read all the comments before replying to any of them.

Looking at the proposed change, I'm not sure there's any value to continuing to support access by string. If it's only to provide backwards compatibility then I don't think we need to worry since the current release is 1.0 in name only.

source/fab/artefacts.py
```diff
@@ -4,35 +4,131 @@
...
+    PREPROCESSED_C = auto()
+    FORTRAN_BUILD_FILES = auto()
+    C_BUILD_FILES = auto()
+    X90_BUILD_FILES = auto()
+    CURRENT_PREBUILDS = auto()
+    PRAGMAD_C = auto()
+    BUILD_TREES = auto()
+    OBJECT_FILES = auto()
+    OBJECT_ARCHIVES = auto()
+    EXECUTABLES = auto()
```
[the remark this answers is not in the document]
hiker: Hmm - typically, the keys/set are used in communicating between tools, a bit like a producer and consumer :) So there would be likely (at least) two tools involved in defining one set.

Yes, we could have functionality where a tool could request a new 'key', which can then be used (making sure no incidental usage of the same key could occur). And that would take care of the string/enum keys we have atm. But then we would need to transfer the keys between various processors:
```
psyclone_file = ArtefactSet.getKey()
```
Then we would need to have this variable when we detect .x90 files, and in PSyclone. Or of course, we could just create a new object that stores a 'set' of something, and provide them as parameters to each step. Not sure tbh, there is some simplicity in the current way, which I think might be a nice advantage for Fab.

source/fab/artefacts.py
```diff
@@ -4,35 +4,131 @@
...
         self.clear()
-        self[CURRENT_PREBUILDS] = set()
+        for artefact in ArtefactSet:
+            if artefact in [ArtefactSet.OBJECT_FILES,
+                            ArtefactSet.OBJECT_ARCHIVES]:
+                # ObjectFiles store a default dictionary (i.e. a non-existing
+                # key will automatically add an empty `set`)
+                self[artefact] = defaultdict(set)
+            else:
+                self[artefact] = set()
```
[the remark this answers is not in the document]
hiker: Oh, I agree, I totally hated that :) My idea (for a different PR): we store two sets:
1. We map OBJECT_FILES (similar for OBJECT_ARCHIVES) to the set of root names (a list of strings)
2. We map each root name (string) to the set of object files required.

That would allow us to remove the special cases, but it will require to use a mixture of enums/strings as keys.

Of course, the trivial solution: each of these two sets just store a single dictionary, but that feels ugly as well. At least then the knowledge about what is stored there is restricted to the producer/consumer only (which already has this knowledge), it's not in ArtefactStore anymore.

source/fab/artefacts.py
```diff
@@ -4,35 +4,131 @@
...
+        '''
+        if isinstance(files, list):
+            files = set(files)
+        elif not isinstance(files, Iterable):
+            # We need to use a list, otherwise each character is added
+            files = set([files])
+
+        self[collection].update(files)
+
+    def update_dict(self, collection: Union[str, ArtefactSet],
```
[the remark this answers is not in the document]
hiker: Agreed. See my suggestions in the previous comment.

source/fab/artefacts.py
```diff
@@ -4,35 +4,131 @@
...
+        :param suffixes: a string or list of strings specifying the
+            suffixes to copy.
+        '''
+        if suffixes:
+            suffixes = [suffixes] if isinstance(suffixes, str) else suffixes
+            self.add(dest, set(suffix_filter(self[source], suffixes)))
+        else:
+            self.add(dest, self[source])
+
+    def replace(self, artefact: Union[str, ArtefactSet],
```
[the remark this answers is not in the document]
hiker: I don't like that idea. One of my use cases is that I need to replace the Fortran source files with a version that has all protected/private attribute/declaration removed. If I can't remove (and of course add) entries to an artefact, I would need to introduce a completely new artefact set (plus I need to make the processors aware of this change). Replacing/removing+adding seems to be the much easier solution for this case (unless you have a good idea :) )

Documentation/source/writing_config.rst
```diff
@@ -79,7 +79,7 @@ Please see the documentation for :func:`~fab.steps.find_source_files.find_source
 including how to exclude certain source code from the build. More grab steps can be found in the :mod:`~fab.steps.grab`
 module.
 
-After the find_source_files step, there will be a collection called ``"all_source"``, in the artefact store.
+After the find_source_files step, there will be a collection called ``"ALL_SOURCE"``, in the artefact store.
```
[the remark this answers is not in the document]
hiker: That's a much better name. I've renamed it everywhere.

Documentation/source/writing_config.rst
```diff
@@ -269,7 +268,42 @@ then creates the executable.
...
+
+ArtefactStore
+=============
+Each build configuration contains an artefact store, containing various
+sets of artefacts. The artefact sets used by Fab are defined in the
+enum :class:`~fab.artefacts.ArtefactSet`. The most important sets are ``FORTRAN_BUILD_FILES``, 
+``C_BUILD_FILES``, which will always contain all known source files that
+will need to be analysed for dependencies, compiled, and linked. All existing
+steps in Fab will make sure to maintain these artefact sets consistently,
+for example, if a ``.F90`` file is preprocessed, the ``.F90`` file in
```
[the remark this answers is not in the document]
hiker: I assume this is for making the artefact sets read-only? Maybe we can discuss this in person in the telco? As I've mentioned elsewhere, I like the ability to update this :)

Documentation/source/writing_config.rst
hiker: Oh, indeed I didn't know that. Yes, that step always felt odd, but I never looked any closer.

run_configs/lfric/lfric_common.py
hiker: Yep, that's why I removed the comment :)

run_configs/lfric/lfric_common.py
hiker: The issue number (330) is just in the next line. And the ticket is still open. I've labelled it for the next release, tbh, it shouldn't be that much work, I will try to get it done.

run_configs/lfric/lfric_common.py
hiker: Indeed.

source/fab/parse/fortran.py
hiker: Done.

source/fab/steps/archive_objects.py
hiker: "Mark up" as in special comments/directives in the source files? Interesting idea. For most Fab use cases that I can think of, zero-config won't really ever be used, so I am kind of neutral here - if LFRic/um/... needs libs, that can be done (with some extensions to Fab I think??) in a Fab script.

source/fab/steps/archive_objects.py
hiker: (fwiw, the .exe was there before, I just reformatted it :) ).
I've replaced all exe with executable.

tests/unit_tests/test_artefacts.py
hiker: Aeh ... I guess because I forgot to remove it when I figured out how to loop over all enums :) I've removed this one test (since in the lines below we would get an exception an ArtefactSet is not in the ArtefactStore).

Documentation/source/writing_config.rst
```diff
@@ -227,14 +226,14 @@ The Analyse step looks for source to analyse in several collections:
 Here we tell the analyser which :term:`Root Symbol` we want to build into an executable.
 Alternatively, we can use the ``find_programs`` flag for Fab to discover and build all programs.
 
-After the Analyse step, there will be a collection called ``"build_trees"``, in the artefact store.
+After the Analyse step, there will be a collection called ``BUILD_TREES``, in the artefact store.
```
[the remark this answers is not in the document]
hiker: See #328.

Documentation/source/writing_config.rst
```diff
@@ -179,7 +179,7 @@ before you run the :func:`~fab.steps.analyse.analyse` step below.
 
 
 After the psyclone step, two new source files will be created for each .x90 file in the `'build_output'` folder.
-These two output files will be added under ``"psyclone_output"`` collection to the artefact store.
+These two output files will be added under ``FORTRAN_BUILD_FILES`` collection to the artefact store.
```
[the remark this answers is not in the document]
hiker: See #328.

hiker:
Thanks for the review. I have renamed ALL_SOURCE->INITIAL_SOURCE as suggested, and opened #328 to find better names for other artefacts. As requested, I've closed the issues that I think are ok now. I'd suggest to leave the read-only-artefactstore for the telco. Not sure about the special handling of dictionary artefacts - open a ticket for later?

hiker:
Oh, I also reformatted some more files to be 80 characters or less (not all of them, some were rather huge, making it hard to see the important changes).

MatthewHambley (approved):
I'm happy that conversations have been started around the things I spotted while reviewing this change. The change itself is a clear step forward and basis for future work.

Issue #297, opened 2024-03-25
Handling of OMP Sentinels


hiker:
Some UM files contains OMP sentinels for use statements, e.g. control/grids/p_to_t_vol.F90:
```
!$ USE compute_chunk_size_mod, ONLY: compute_chunk_size  ! Note OpenMP sentinel
```
Since fparser does not support this, FAB misses this dependency (and compilation fails). While I have a patch more or less ready to go (https://github.com/hiker/fab/tree/13_omp_sentinels), I have discussed this in our telco, and we agreed that this would be a useful feature for fparser to support, see https://github.com/stfc/fparser/issues/443. So I'll add this support to fparser, and once this has been released, we just need to enable this feature in FAB (it will be disabled by default in order not to break compatibility).

hiker:
I have just submitted https://github.com/stfc/fparser/pull/444 to get support for omp conditional sentinels into fparser. I have confirmed that fab now detects the dependencies in the original case when setting the fparser flag to handle them:
```
diff --git a/source/fab/parse/fortran_common.py b/source/fab/parse/fortran_common.py
index 0ed4f3f..8b987d8 100644
--- a/source/fab/parse/fortran_common.py
+++ b/source/fab/parse/fortran_common.py
@@ -142,7 +142,8 @@ class FortranAnalyserBase(ABC):
 
     def _parse_file(self, fpath):
         """Get a node tree from a fortran file."""
-        reader = FortranFileReader(str(fpath), ignore_comments=False)
+        reader = FortranFileReader(str(fpath), ignore_comments=False,
+                                   omp_sentinel=True)
         reader.exit_on_error = False  # don't call sys.exit, it messes up the multi-processing
 
         try:
```

This obviously needs a new fparser release

hiker:
https://github.com/stfc/fparser/pull/444.  has been merged, so once we do the next fparser release we can support these statements. As part of the improved compiler support I am also improving the handling of openmp, so FAB will 'know' whether openmp is enabled or not.

hiker:
Fixed recently.

PR #326, opened 2024-07-17
297 omp sentinels

hiker:
Adds some support for OMP sentinels (#297), which are required for compiling lfric_atm.

The next release of fparser adds support for sentinels to fparser, so this change can then be removed again (#todo in place to remind us, so I would also leave #297 open), we only need to trigger this behaviour in fparser (which means it will not only work for declaration statements). We need this patch for now to continue LFRic compilation (but the source code in question is UM, so I added label UM, and not LFRic).

### DIFF
--- a/source/fab/parse/fortran.py
+++ b/source/fab/parse/fortran.py
@@ -288,7 +288,7 @@ class FortranAnalyser(FortranAnalyserBase):
         comment = obj.items[0].strip()
         if depends_str in comment:
             self.depends_on_comment_found = True
-            dep = obj.items[0].split(depends_str)[-1].strip()
+            dep = comment.split(depends_str)[-1].strip()
             # with .o means a c file
             if dep.endswith(".o"):
                 analysed_file.mo_commented_file_deps.add(dep.replace(".o", ".c"))
--- a/source/fab/parse/fortran.py
+++ b/source/fab/parse/fortran.py
@@ -11,6 +11,7 @@ import logging
 from pathlib import Path
 from typing import Union, Optional, Iterable, Dict, Any, Set
 
+from fparser.common.readfortran import FortranStringReader
 from fparser.two.Fortran2003 import (  # type: ignore
     Entity_Decl_List, Use_Stmt, Module_Stmt, Program_Stmt, Subroutine_Stmt, Function_Stmt, Language_Binding_Spec,
     Char_Literal_Constant, Interface_Block, Name, Comment, Module, Call_Stmt, Derived_Type_Def, Derived_Type_Stmt,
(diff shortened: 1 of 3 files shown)


tests/unit_tests/parse/fortran/test_fortran_analyser.f90
```diff
@@ -17,6 +17,11 @@ SUBROUTINE internal_sub
         RETURN
     END SUBROUTINE internal_sub
 
+    SUBROUTINE openmp_sentinel
+!$ USE compute_chunk_size_mod, ONLY: compute_chunk_size  ! Note OpenMP sentinel
+!$ USE test that is not a sentinel with a use statement inside
+    END SUBROUTINE openmp_sentinel
```
MatthewHambley, line 23: How about testing some other directives such as `!GCC$ unroll 6`, `!DIR$ assume (mod(p, 6) == 0)`, `!$omp do` and `!$acc parallel copyin (array, scalar)`.

source/fab/parse/fortran.py
```diff
@@ -284,15 +285,38 @@ def _process_comment(self, analysed_file, obj):
...
             if dep.endswith(".o"):
                 analysed_file.mo_commented_file_deps.add(dep.replace(".o", ".c"))
             # without .o means a fortran symbol
             else:
                 analysed_file.add_symbol_dep(dep)
+        if comment[:2] == "!$":
+            # Check if it is a use statement with an OpenMP sentinel:
+            # Use fparser's string reader to discard potential comment
+            # TODO #13: once fparser supports reading the sentinels,
+            # this can be removed.
```
MatthewHambley, line 302: Will there be a problem if this code is used with an fparser with the directive support implemented?

MatthewHambley (changes requested):
What's the timescale for the fParser fix and the need for directive support in Fab?

source/fab/parse/fortran.py
hiker: No. If we enable the handling of omp sentinels, no comment will be returned from fparser (instead the proper Fortran tree will be used).

The only inconsistent behaviour would be if we do not enable sentinels in fparser, but leave this code in I'd guess. Note that my intention is that the config object will be aware if the build is with or without openmp (since omp flags are compiler-specific, so this way the config object can trigger adding the appropriate flag in the compiler tools - either enable or disable openmp, e.g. Cray has openmp enabled by default).

tests/unit_tests/parse/fortran/test_fortran_analyser.f90
hiker: Done (though since these directives are ignored, there isn't much I can explicitly compare, except no new dependencies are added, so all that changed in the tests were the checksums)

hiker:
Ready for next review.

hiker:
> What's the timescale for the fParser fix and the need for directive support in Fab?

I've just submitted my last additional PR that I wanted for the next release, and there seems to be only one issue that might need to be handled before the next release. An fparser release is usually done very quickly, so I would hope (depending on priorities) to get this in maybe a month? 

I originally intended not to push this PR to the official repo at all, but we had to integrate all these changes into our PRs (it became very messy to keep track and merging things), so it simplifies things a bit for us. E.g. #301 has this patch and it's not easy to remove it (conflicts).

MatthewHambley (approved):
I've raised #327 to remove these changes when they are no longer needed.

Issue #318, opened 2024-06-20
Properly handle compiler versions


hiker:
Atm compiler versions are only handled as strings. We need to properly convert them to integer-tuples to allow comparison, and also support compiler-specific compiler flags to be added

jasonjunweilyu:
This is the current way of how LFRic makefiles do compiler version comparison. I have converted this into Python:

 ```
compiler_version_comparison = ''.join(f"{int(version_component):02d}" for version_component in compiler.get_version().split('.')) 

if compiler_version_comparison >= '190000':
     runtime_flags = ...
else: ...
```

So current `get_version()` returns a string `1.20.0`. It is converted to string `012000` and then compared with string `190000`

hiker:
Has been fixed quite some time ago.

PR #334, opened 2024-08-14
Properly handle compiler versions

lukehoffmann:
This PR covers issue #318

`Compiler.get_version()` will return a tuple of integer values, e.g. `(8, 5, 0)` instead of the string `"8.5.0"`. This makes it possible to do greater-than and lesser-than comparisons with required compiler versions.

The different compiler classes will also verify that the version output contains the expected compiler name.


source/fab/tools/compiler.py
```diff
@@ -92,68 +93,95 @@ def check_available(self) -> bool:
...
+            return True
+        except RuntimeError as err:
+            # Compiler does not exist, or version could not be handled:
+            self.logger.error(f'Error getting compiler version: {err}')
             return False
 
-        # An empty string is returned if some other error occurred when trying
-        # to get the compiler version.
-        return version != ""
-
```
allynt, line 103: I'm not very familiar w/ this code - is there a chance an exception other than `RuntimeError` could happen?  In that case should you remove the return statement in the try block and re-insert the `version != ""` here?

source/fab/tools/compiler.py
```diff
@@ -92,68 +93,95 @@ def check_available(self) -> bool:
...
+            raise RuntimeError(f"Unexpected version output format for "
+                               f"compiler '{self.name}'. Should be numeric "
+                               f"<n.n[.n, ...]>: {version_string}") from err
+
+        # Expect at least 2 integer components, i.e. major.minor[.patch, ...]
+        if len(version) < 2:
+            raise RuntimeError(f"Unexpected version output format for "
+                               f"compiler '{self.name}'. Should have at least "
+                               f"two parts, <n.n[.n, ...]>: {version_string}")
+
```
allynt, line 139: Why is this changed from a log warning to a runtime error?

lukehoffmann: I didn't mention in the summary, but part of this change is the design decision that `get_version()` should only return a valid version tuple. if it can't determine a version, it raises a runtime error.
Previously it could return an empty string in some cases, but we thought that just made things unclear, and increased the amount of checking you might need to do when calling `get_version()`.

source/fab/tools/compiler.py
lukehoffmann: `get_version()` will no longer return an empty string, and there is no other 'empty' type value to check for. It will always raise a `RuntimeError` if it can't determine the version.
`RuntimeError` is the only expected error type from the underlying `self.run()` function (which is in `tool.py`). It handles the `FileNotFoundError` if the compiler isn't found at all.

source/fab/tools/compiler.py
hiker: Agreed - if we can't detect the version number (be it because the compiler has changed, or we have a bug in Fab), we need to abort. In LFRic we will have compiler flags added depending on compiler version, so it is important that we get the version correct.

allynt (approved):
All my comments have been addressed.  So I am approving this.

PR #339, opened 2024-10-12
Recovering anything of value from the old tests

MatthewHambley:
There's potentially some useful stuff in these old tests but the new implementation is so completely different (and weird)  that I have up after a few scraps were saved.


allynt (approved):
Looks good to me.

PR #352, opened 2024-11-20
Compiler wrapper

hiker:
Introduces proper compiler wrapper, e.g. for mpif90.


source/fab/tools/compiler_wrapper.py
```diff
@@ -0,0 +1,206 @@
...
+        orig_compiler_name = self._compiler.exec_name
+        self._compiler.change_exec_name(self.exec_name)
+        if add_flags is None:
+            add_flags = []
+        if self._compiler.category is Category.FORTRAN_COMPILER:
+            # Mypy complains that self._compiler does not take the syntax
+            # only parameter. Since we know it's a FortranCompiler.
+            # do a cast to tell mypy that this is now a Fortran compiler
+            # (or a CompilerWrapper in case of nested CompilerWrappers,
+            # which also supports the syntax_only flag anyway)
```
MatthewHambley, line 163: This suggests there might be a problem somewhere. Is there an issue to investigate?

source/fab/tools/compiler_wrapper.py
```diff
@@ -0,0 +1,206 @@
...
+    It will be using the name "mpif90-COMPILER_NAME" and calls `mpif90`.
+    All flags from the original compiler will be used when using the wrapper
+    as compiler.
+
+    :param compiler: the compiler that the mpif90 wrapper will use.
+    '''
+
+    def __init__(self, compiler: Compiler):
+        super().__init__(name=f"mpif90-{compiler.name}",
+                         exec_name="mpif90", compiler=compiler, mpi=True)
```
MatthewHambley, line 191: I'm concerned that we end up with a combinatorial problem here. For example when we start thinking about profilers, some of which wrap the compiler, we could end up with objects for compiler, MPI compiler, profiled compiler and profiled MPI compiler.

Can we wrap the compiler dynamically, rather than holding it named in the tool kit, and perform a test in the constructor to ensure there is actually an MPI built with that compiler at that point?

source/fab/tools/tool.py
```diff
@@ -91,11 +91,26 @@ def exec_name(self) -> str:
         ''':returns: the name of the executable.'''
         return self._exec_name
 
+    def change_exec_name(self, exec_name: str):
```
MatthewHambley, line 94: I'm concerned that this is open to abuse or mistaken use. Would it be better to provide a method which returns the command line without the executable which a wrapping class could then use to prepend with its own executable and arguments?

tests/unit_tests/tools/test_compiler_wrapper.py
```diff
@@ -0,0 +1,348 @@
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+
+'''Tests the compiler wrapper implementation.
+'''
+
+from pathlib import Path, PosixPath
```
MatthewHambley, line 10: I've never found I need to explicitly import `PosixPath`. For instance  `cwd = Path()` should give the currently selected directory.

MatthewHambley (changes requested):
I've repeated the issue I raised by e-mail so you can post your response here for posterity.

tests/unit_tests/tools/test_compiler_wrapper.py
hiker: Good point.

source/fab/tools/tool.py
hiker: After some thinking, I have added a #TODO for #370. Main reason is that I nearly did exactly that when allowing linker to be 'wrapped'. Since I might have modified related functions, I'll address this issue as part of the branch `linker_wrapper_new` to avoid fighting with conflicts (it's around 6 PRs ahead of this).

MatthewHambley (approved):
Lets move things forward a bit and get to the next change.

PR #368, opened 2024-11-29
Add shell tool clean

hiker:
This just adds a shell tool. The tool itself is not used directly (though it could potentially replace some common tools? Not sure that we gain anything from it though :) ).

I am adding this since it is an extremely useful tool for Fab scripts. For example, many scripts might want to query `nf-config` for compiler- and linker flags for using NetCDF. Since it was decided not to support running Fab-triggered command using a shell (which would allow a configuration to use `$(nf-config ..fflags)` etc), applications now need to explicitly run a shell command.
With a shell tool in Fab (which will check various shells to find one that is available on the platform), this becomes a lot more convenient and reduces code duplication.

Unfortunately, I could not find a generic way of testing if a shell is available with a single parameter (e.g. some shells have no version command), so the most portable way I could think of was to run an `echo something` command ... which then requires at least two parameters (`-c ` and `echo something`), requiring many changes in tool to support a list of strings instead of only a single string.


source/fab/tools/shell.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+
+from pathlib import Path
+from typing import List, Union
+
+from fab.tools.category import Category
+from fab.tools.tool import Tool
+
+
+class Shell(Tool):
```
MatthewHambley, line 18: The general rule for scripting is to only ever use the Posix standard `sh`. That shell is always present on any Posix or Posix-like system and is always called and used the same way. This greatly improves interoperability.

As such I would recommend following suite and not trying to guess if a shell is present but to always use `sh`.

This may also mean the change to the `availability_option` type is unnecessary.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html

source/fab/tools/tool.py
```diff
@@ -63,7 +63,8 @@ def check_available(self) -> bool:
         :returns: whether the tool is working (True) or not.
         '''
         try:
-            self.run(self._availability_option)
+            op = self._availability_option
+            self.run(op)
```
MatthewHambley, line 67: Why has an intermediate variable been added?

source/fab/tools/tool_repository.py
```diff
@@ -72,6 +72,12 @@ def __init__(self):
                     Ar, Fcm, Git, Psyclone, Rsync, Subversion]:
             self.add_tool(cls())
 
+        # Add the common shells. While Fab itself does not need this,
+        # it is a very convenient tool for user configuration (e.g. to
+        # query nc-config etc)
+        for shell_name in ["bash", "sh", "ksh", "dash"]:
+            self.add_tool(Shell(shell_name))
```
MatthewHambley, line 79: See previous comment about supporting non-standard shells.

source/fab/tools/tool_repository.py
```diff
@@ -176,7 +184,12 @@ def get_default(self, category: Category,
 
         # If not a compiler or linker, return the first tool
         if not category.is_compiler and category != Category.LINKER:
-            return self[category][0]
+            for tool in self[category]:
+                if tool.is_available:
+                    return tool
+            tool_names = ",".join(i.name for i in self[category])
+            raise RuntimeError(f"Can't find available '{category}' tool. "
+                               f"Tools are '{tool_names}'.")
```
MatthewHambley, line 192: This is new behaviour extraneous to  the description. Please add it to the description.

tests/unit_tests/steps/test_grab.py
```diff
@@ -15,24 +19,37 @@
...
                                       tool_box=ToolBox())
+        # Since is_available calls run, in order to test a single run call,
+        # we patch is_available to be always true.
         with mock.patch('pathlib.Path.mkdir'):
             with mock.patch('fab.tools.tool.Tool.run') as mock_run:
-                grab_folder(mock_config, src=grab_src, dst_label=dst)
+                with mock.patch(
+                        'fab.tools.tool.Tool.is_available',
+                        new_callable=mock.PropertyMock) as is_available:
+                    is_available.return_value = True
```
MatthewHambley, line 51: This may not be necessary if we support only Posix shell.

tests/unit_tests/tools/test_shell.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # Test behaviour if a runtime error happens:
+    with mock.patch("fab.tools.tool.Tool.run",
+                    side_effect=RuntimeError("")) as tool_run:
+        assert not bash.check_available()
+
+
+def test_shell_exec_single_arg():
+    '''Test running a shell script without additional parameters.'''
+    bash = Shell("ksh")
```
MatthewHambley, line 42: It may not matter if we only support Posix shell, but why is the KSH shell in a variable called `bash`?

MatthewHambley (changes requested):
The key issue with this change is whether we should try to support multiple shells or just the Posix standard shell. There is potential to simplify the change if we support only the standard shell and I'm not sure we get any advantage for supporting others.

source/fab/tools/shell.py
hiker: I clearly remember the UKMO having scripts with specific ksh-only syntax :)  So I would suggest that we try to support different shells, especially since it's basically no work. I will change the order in which shells are creates, so `sh` becomes the default.

The support for multiple parameters when checking availability was actually added for `sh`, since it does not support any version command :(

source/fab/tools/tool.py
hiker: Absolutely no idea ;) Maybe I had some debug print in between? I've reverted that.

source/fab/tools/tool_repository.py
hiker: Order changed to have `sh` first.

source/fab/tools/tool_repository.py
hiker: Exception description added, also added a sentence or two to the docs.

tests/unit_tests/steps/test_grab.py
hiker: I think we should still try if even `sh` is available - who knows ;) And the advantage is that we can then use `assert_called_once_with` (instead of having to check for two calls).

tests/unit_tests/tools/test_shell.py
hiker: Because ... I used the very advanced and agile approach of copy&paste :)   Fixed, thanks!

hiker:
@MatthewHambley, thanks for your comments. Notice that this PR is a bit in 'your future', it's 4 prs ahead of main, which is why I had it marked for Jason ;)

Not sure how you would like to proceed. I won't try to merge the PRs in out-of-order - I've spent way too much time when I migrated our branches across (admittedly because of user error ... I mixed up the order of two branches ;) ). So you can either continue, and once this is OK, I'll keep it in its position, and we just merge it when we reach it. Or you wait till I 'trickle' this PR through to you as the next one, and I'll let Jason finish this for now (i.e. for a merge into develop)

hiker:
Ready for next review.

source/fab/tools/shell.py
MatthewHambley: That certainly happened and it was just as wrong then as it is now.

Clearly there is work to support multiple shells (and maintain that support) which we can avoid by not doing it.

As far as I can see fetching the version is only a means to determine if the shell is present. But `sh` is **always** present on Posix and Posix-like systems, so there is no need to check.

hiker: > That certainly happened and it was just as wrong then as it is now.

No comment :)

I just accept that these specific scripts do exist.

> 
> Clearly there is work to support multiple shells (and maintain that support) which we can avoid by not doing it.
Not really, it's just a simple loop to create three different instances (and it is actually `sh` that caused all the additional coding overhead).


> 
> As far as I can see fetching the version is only a means to determine if the shell is present. But `sh` is **always** present on Posix and Posix-like systems, so there is no need to check.

I am not fetching a version number (I would if I could :) ), I am running 'echo hello' (or so). I don't know if sh is always present (I leave the posix-like out here ;) ). I just don't see what can hurt if we add two more shells in to cover more cases (and making it easy for any site to add `this-is-the-one-and-only-shell` to it ;)

Plus, I prefer all tools to be handled the same. Just assuming that `sh` always works seems wrong, any other tool also first checks if it can run some dummy command, which means we can get better feedback to the user.

jasonjunweilyu (changes requested):
Hi @hiker , this PR is fine for me. I only have a minor comment. I think this addition of shell tool adds an additional check for user's environment and may help at some stage for signalling incorrect environment configurations.

source/fab/tools/tool.py
```diff
@@ -139,7 +139,7 @@ def __str__(self):
 
     def run(self,
             additional_parameters: Optional[
-                Union[str, List[Union[Path, str]]]] = None,
+                Union[str, Sequence[Union[Path, str]]]] = None,
```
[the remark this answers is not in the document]
hiker: Not sure what you mean. If I use `List`, mypy complains:
```
tools/tool.py:66: error: Argument 1 to "run" of "Tool" has incompatible type "str | list[str]"; expected "str | list[Path | str] | None"  [arg-type]
```
Line 66 is:
```
            self.run(self._availability_option)
```
with `self._availability_option` getting assigned a value of type:
```
Union[str, List[str]]
```

I think I understand the reason: the `run` method might modify the list and add an element of type `Path` to it, since the argument is declared to include `Path`, but this would then not be compatible with `List[str]` anymore. I assume a generic Sequence does not allow modifications (??) - hence it can't add a variable of a different type into the parent list??

A string might be a sequence, but that would indicate that we would iterate over each character I think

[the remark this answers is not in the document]
jasonjunweilyu: I was wondering whether `Sequence` includes `string`, and why you changed from list to sequence. Thanks for the explanation.

jasonjunweilyu (approved):
I think this is now ready to be merged into `develop` branch. The shell tool addition may help with debugging for user environment when something goes really wrong.

source/fab/tools/shell.py
MatthewHambley: The reason for the link I sent was to show you the part of the Posix standard which *requires* the presence and capabilities of `sh`.

The reason supporting multiple shells hurts is because people might use different shells. What then happens if I try to run the build system on a platform which doesn't have that shell? It doesn't work is what happens.

This is why shell scripts should always be written using `sh`, the only standard shell. We can expect target platforms to have `sh` we can't expect them to have `bash`, `ksh` or `zsh`.

However I can see you have your heart set on this and frankly it isn't a hill worth dying on. We can always refactor it later if need be.

MatthewHambley (approved):
I approve this change under protest.

hiker:
I have checked, all applications we have at are de-facto sh script, so since you are really concerned, we can remove bash etc from the loop, so only sh gets created. Note that I still want the test that sh works, just to be consistent among all tools, and give better error messages to the user (i.e. 'sh' is not available, as opposed a potential cryptic error message when the user tries to run a script, which then might not work).
Thanks!

source/fab/tools/shell.py
hiker: > The reason supporting multiple shells hurts is because people might use different shells. What then happens if I try to run the build system on a platform which doesn't have that shell? It doesn't work is what happens.

This shell tool is not used at all by Fab, it is meant as a convenience for site-specific scripts. I've added this script since you didn't like the support of running compilers in a shell. This implies that  a site cannot use `$(nf-config -fflags)` to get for example the NetCDF flags, which we had assumed (other scripts besides nf-config are possible of course, including potentially using pkg-config this way - but that's another discussion

Therefore, site-specific scripts now need to run this command as a shell separately to detect for example the netcdf flags easily. So there is absolutely no risk that a script might not work for other sites, since other-sites won't use the same site-specific setup.

This is under the assumption that we will make sure that Fab does not add any bash/ksh/...-only scripts, which I intend to do.

PR #372, opened 2024-12-11
#371 Fix incorrect number of files reported when two stage compilation is done

### LINKED ISSUE #371: Two-stage compile has 0 file compiled at the second stage
When two-stage compilation is enabled, the build failed for LFRic gungho at the moment with an error `/scratch/hc46/hc46_gitlab/lfric_fab/gungho_model-mpif90-ifort/build_output/_prebuild/physics_mappings_alg_mod.cb6a16b9a.o` not found at link stage.

It is observed that the following messges are logged during the compile_fortran step:

`Starting two-stage compile: mod files, multiple passes`
...
`Finalising two-stage compile: object files, single pass`
...
`stage 2 compiled 0 files`

There are actually a number of compiling commands being executed between `Finalising two-stage compile: object files, single pass` and `stage 2 compiled 0 files`. So I am wondering whether there is a bug in updating the build tree. 

The build log is here: https://git.nci.org.au/bom/ngm/lfric/lfric_atm-fab/-/jobs/80826

hiker:
Fixed #371 (the old code assumed that the result was a list of CompiledFiles, but the returned value was a list of tuples, so the filtering by type always returned 0).

### DIFF
--- a/source/fab/steps/compile_fortran.py
+++ b/source/fab/steps/compile_fortran.py
@@ -111,7 +111,8 @@ def compile_fortran(config: BuildConfig,
     log_or_dot_finish(logger)
 
     if syntax_only:
-        logger.info("Finalising two-stage compile: object files, single pass")
+        logger.info(f"Finalising two-stage compile: object files, single "
+                    f"pass for {len(uncompiled)} files.")
         mp_common_args.syntax_only = False
 
         # A single pass should now compile all the object files in one go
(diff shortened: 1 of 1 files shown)


hiker:
@MatthewHambley, this is directly for you to merge into main, it's very small and does not affect any code deveopment we are doing - so no sense to bring this into develop first.

hiker:
I'm taking this back to include this with a complete fix for #371 (i.e. including the race condition causing failures).

PR #376, opened 2025-01-17
Update clang binding package

MatthewHambley:
It looks like the clang binding package has changed its name. We should adopt the new name.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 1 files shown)


hiker:
According to https://pypi.org/project/clang/ and https://pypi.org/project/libclang/, the first one was never the official name.

hiker (approved):
Ready to be merged, thanks for the update.

hiker:
I am not authorized to merge ... sorry. Could you perhaps get that sorted out?

PR #380, opened 2025-02-04
More linker wrapper improvements

hiker:
Make sure linkers will test if the corresponding compiler actually exists (plus a minor bug fix).


hiker:
Wrong branch :(

PR #384, opened 2025-02-05
Additional compilers clean

hiker:
Removes support for "-" in nvidia compiler versions (e.g. 2.5-0). This caused problems with compiler wrapper:
`nvfortran` would accept 2.5-0, but `mpif90(nvfortran)` would not accept (resulting in `mpif90` not available for nvfortran).

Fixing this properly is a larger effort, while only supporting non-development version numbers (according to semantic versioning) is easy and is all we need.


hiker:
Changes are on small_fixes branch instead.

Issue #341, opened 2024-10-21
Remove work-around for (previously unsupported) string concatenation in fparser


hiker:
I've fixed the issue with a `stop` statement that contains a string concatenation in current fparser. Could you provide me the source file that had this problem so I can verify the fix?

I grep'ed through the lfric source (for `stop.*//`, and couldn't find anything).

hiker:
I checked the run_configs, and found which file was causing the problem - feign_config_mod.f90, which is automatically created.

The latest version of this file does not contain a stop with a concatenation anymore, so we can't verify this. But it is tested in fparser. 

So I'll change this ticket to just remove the work around code that is not necessary anymore from the run_configs examples.

hiker:
Affected files:
```
fab/run_configs$ grep fparser_workaround_stop_concatenation -r .
./lfric/mesh_tools.py:from lfric_common import API, configurator, fparser_workaround_stop_concatenation
./lfric/mesh_tools.py:        fparser_workaround_stop_concatenation(state)
./lfric/gungho.py:from lfric_common import (API, configurator, fparser_workaround_stop_concatenation,
./lfric/gungho.py:        fparser_workaround_stop_concatenation(state)
./lfric/atm.py:from lfric_common import (API, configurator, fparser_workaround_stop_concatenation,
./lfric/atm.py:        fparser_workaround_stop_concatenation(state)
./lfric/lfric_common.py:def fparser_workaround_stop_concatenation(config):
```

hiker:
Fixed in #409.

PR #386, opened 2025-02-07
Removed fparser stop concatenation workround after fparser update

jasonjunweilyu:
This is related to Issue #341 . The work around for stop concaenation is no longer needed after fparser update. This PR removes this redundant step from the example configs.


hiker (approved):
Great, thanks a lot. Changes are all clean.

PR #389, opened 2025-02-14
Bring develop up to main

hiker:
This PR is only to bring develop up to master. I realised that when one older PR got merged into main, we've updated the individual PRs that follow, but we never updated develop, so develop misses all the changes asked for in the code review on master.

This should bring the latest changes from main into master.

I'd guess in the future when a new PR is merged into master, we might have to:
1. bring all follow-up PRs up to date (by merging main into branch1, branch1 into branch2, ... (which is what we have been doing)
2. merge at least branch1 into develop again (and if there should be conflicts in further branches that were resolved in step 1, I guess we merge this all back in.

### DIFF
--- a/Documentation/source/advanced_config.rst
+++ b/Documentation/source/advanced_config.rst
@@ -186,20 +186,75 @@ Linker flags
 ------------
 
 Probably the most common instance of the need to pass additional arguments is
-to specify 3rd party libraries at the link stage.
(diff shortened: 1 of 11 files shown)


jasonjunweilyu (approved):
Just realized that I still have this PR for review. Sorry for forgetting about it. I think the changes are ready to be in `develop`.

PR #391, opened 2025-02-28
Adopt subprocess fixture for testing.

MatthewHambley:
Rather than variations of the same hand-mocking of shelling out to a tool this change takes advantage of an existing fixture.

It also pulls in a fixture to create a fake filing system in memory which should be faster and less prone to faults based on what the developer has in their file tree.


tests/__init__.py
hiker, line 1: Is that file required?

tests/conftest.py
```diff
@@ -1,70 +1,129 @@
...
+
+from pytest import fixture
+from pytest_subprocess.fake_process import FakeProcess, ProcessRecorder
+
+from fab.tools.compiler import CCompiler, FortranCompiler
+from fab.tools.linker import Linker
+from fab.tools.tool_box import ToolBox
+
+
+def not_found_callback(process):
```
hiker, line 19: Could you please add some comments to all these functions.

tests/system_tests/psyclone/test_psyclone_system_test.py
```diff
@@ -170,42 +176,44 @@ def test_run(self, config, psyclone_lfric_api):
...
+    def test_transformation_script(self) -> None:
         psyclone_tool = Psyclone()
         psyclone_tool._version = (2, 4, 0)
         psyclone_tool._is_available = True
 
         mock_transformation_script = mock.Mock(return_value=__file__)
         with mock.patch('fab.tools.psyclone.Psyclone.run') as mock_run_command:
             mock_transformation_script.return_value = Path(__file__)
-            psyclone_tool.process(api=psyclone_lfric_api,
+            psyclone_tool.process(api='dynamo0.3',
```
hiker, line 216: While I am ok with removing the `psyclone_lfric_api` definintion (which allowed them all to be changed at one location), you then need to switch to `lfric`, not `dynamo0.3`, which is deprecated.

tests/system_tests/psyclone/test_psyclone_system_test.py
```diff
@@ -219,7 +227,7 @@ def test_transformation_script(self, psyclone_lfric_api):
             mock_transformation_script.assert_called_once_with(Path(__file__), None)
             # check transformation_script is passed to psyclone command with '-s'
             mock_run_command.assert_called_with(
-                additional_parameters=['-api', psyclone_lfric_api,
+                additional_parameters=['-api', 'dynamo0.3',
```
hiker, line 230: lfric

tests/unit_tests/steps/test_compile_c.py
```diff
@@ -50,110 +59,138 @@ def test_compile_c_wrong_compiler(content):
...
     # C compiler
-    mp_common_args = mock.Mock(config=config)
-    with pytest.raises(RuntimeError) as err:
-        _compile_file((None, mp_common_args))
-    assert ("Unexpected tool 'mock_c_compiler' of category "
-            "'FORTRAN_COMPILER' instead of CCompiler" in str(err.value))
+    mp_common_args = Mock(config=config)
+    with raises(RuntimeError) as err:
+        _compile_file((Mock(), mp_common_args))
+    assert str(err.value) == "Unexpected tool 'some C compiler' of category " \
```
hiker, line 65: pep8 (https://peps.python.org/pep-0008/#maximum-line-length):
"Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation."

tests/unit_tests/steps/test_compile_c.py
```diff
@@ -50,110 +59,138 @@ def test_compile_c_wrong_compiler(content):
...
-        # ensure it sent a metric from the child process
-        send_metric.assert_called_once()
+    def test_vanilla(self, content,
+                     fake_process: FakeProcess, monkeypatch) -> None:
+        """
+        Tests correct use of compiler.
+        """
+        config, _ = content
+
+        monkeypatch.setenv('CFLAGS', '-Denv_flag')
```
hiker, line 80: Oh, I didn't know that - that's neat :)

hiker (commented):
Still more to come, sorry

tests/system_tests/zero_config/test_zero_config.py
```diff
@@ -1,53 +1,76 @@
...
+    def test_c(self, tmp_path: Path) -> None:
+        """
+        Tests a sample C project.
 
-    def test_fortran_explicit_gfortran(self, tmp_path):
-        # test the sample project in the fortran dependencies system test
-        kwargs = {'project_label': 'fortran explicit gfortran', 'fab_workspace': tmp_path, 'multiprocessing': False}
+        ToDo: Fragility due to assumption of source donor.
+        """
+        pytest.importorskip('clang', reason="Missing libclang bindings.")
```
hiker, line 44: Nice!

tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -19,81 +18,82 @@
 
 
 class TestArchiveObjects:
-    '''Test the achive step.
-    '''
+    """
+    Tests the archive step.
+    """
+    def test_for_exes(self, stub_tool_box: ToolBox,
+                      fs, fake_process: FakeProcess) -> None:
```
hiker, line 25: Not your fault of course - but that is a HORRIBLE name. Took me forever to find out what on earth `fs` might even be. Luckily there was one test that typed it :)

hiker (changes requested):
OK, I am stopping here and pass it back to you (extra long weekend) - main reason: could you explain to me the advantage of using `fs`? I have the feeling I am missing something :) So basically, it is a temporary directory, that is just automatically cleared at the end of a test. In my (PSyclone-heavy ;) ) experience, I often use the files in tmp to debug failing tests. So if files are instead created in tmp, I can't debug them to see e.g. what might be wrong. We also have a fixture for changing into tmpdir, which seems to be the equivalent of fs, except in case of failures you can analyse the files. I would also prefer slightly smaller PRs ;) I am aware that this is already a split of an even larger one, but I am sure I missed a `fs` fixture a couple of times before wondering what that might be ;) I'll  pass it back to you at least for clarification of `fs`, and pick it up next week.

hiker:
Oh, also bringing it up to recent `develop` would be great - the psyclone changes in master are causing the conflict ;)

source/fab/artefacts.py
```diff
@@ -82,7 +82,7 @@ def add(self, collection: Union[str, ArtefactSet],
         self[collection].update(files)
 
     def update_dict(self, collection: Union[str, ArtefactSet],
-                    key: str, values: Union[str, Iterable]):
+                    key: Optional[str], values: Union[str, Iterable]):
         '''For ArtefactSets that are a dictionary of sets: update
```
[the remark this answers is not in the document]
MatthewHambley: Good spot but the `None` value is actually used for something. It identifies the situation where there are no executables. i.e. libraries.

source/fab/dep_tree.py
```diff
@@ -98,7 +98,7 @@ def from_dict(cls, d):
         return result
 
 
-def extract_sub_tree(source_tree: Dict[Path, AnalysedDependent], root: Path, verbose=False)\
+def extract_sub_tree(source_tree: Mapping[Path, AnalysedDependent], root: Path, verbose=False)\
```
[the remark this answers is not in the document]
MatthewHambley: As the highest level abstract of a dictionary type object the advantage is that we can accept anything of this type.

Since it is the mapping nature we need (i.e. the ability to set and retrieve values based on keys) it is conceptually the correct option.

However in reality we are unlikely to use anything which does not derive from `Dict` plus we use it all over the place and this is the only place where I've changed it. As such I have reverted it. A move to `Mapping` can happen (or not) as a different ticket.

source/fab/tools/compiler.py
```diff
@@ -65,11 +66,12 @@ def __init__(self, name: str,
         self._output_flag = output_flag if output_flag else "-o"
         self._openmp_flag = openmp_flag if openmp_flag else ""
         self.add_flags(os.getenv("FFLAGS", "").split())
+        self.__version_argument = version_argument or '--version'
         self._version_regex = version_regex
 
     @property
     def mpi(self) -> bool:
-        ''':returns: whether this compiler supports MPI or not.'''
+        '''Returns whether this compiler supports MPI or not.'''
```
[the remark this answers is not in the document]
MatthewHambley: No idea, I've put it back.

source/fab/tools/compiler.py
```diff
@@ -82,7 +84,7 @@ def openmp(self) -> bool:
 
     @property
     def openmp_flag(self) -> str:
-        ''':returns: the flag to enable OpenMP.'''
+        '''Returns the flag to enable OpenMP.'''
```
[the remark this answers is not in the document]
MatthewHambley: ditto

source/fab/tools/compiler_wrapper.py
```diff
@@ -163,15 +162,15 @@ def compile_file(self, input_file: Path,
             # which also supports the syntax_only flag anyway)
             self._compiler = cast(FortranCompiler, self._compiler)
             self._compiler.compile_file(input_file, output_file, openmp=openmp,
-                                        add_flags=self.flags + add_flags,
+                                        add_flags=self._flags + add_flags,
```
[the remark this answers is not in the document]
MatthewHambley: No problem.

source/fab/tools/compiler_wrapper.py
```diff
@@ -163,15 +162,15 @@ def compile_file(self, input_file: Path,
...
                                         syntax_only=syntax_only,
                                         )
         else:
             if syntax_only is not None:
                 raise RuntimeError(f"Syntax-only cannot be used with compiler "
                                    f"'{self.name}'.")
             self._compiler.compile_file(input_file, output_file, openmp=openmp,
-                                        add_flags=self.flags+add_flags
+                                        add_flags=self._flags+add_flags
```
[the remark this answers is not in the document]
MatthewHambley: As per previous comment.

source/fab/tools/tool.py
```diff
@@ -182,7 +182,7 @@ def run(self,
             res = subprocess.run(command, capture_output=capture_output,
                                  env=env, cwd=cwd, check=False)
         except FileNotFoundError as err:
-            raise RuntimeError(f"Command '{command}' could not be "
+            raise RuntimeError(f"Command '{' '.join(command)}' could not be "
```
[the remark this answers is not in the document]
MatthewHambley: A good point.

tests/__init__.py
MatthewHambley: If you want to be able to run all tests with `pytest tests` it is.

tests/conftest.py
MatthewHambley: Of course.

tests/system_tests/psyclone/test_psyclone_system_test.py
MatthewHambley: Done

tests/unit_tests/steps/test_compile_c.py
MatthewHambley: Not sure I agree with that but it is the standard.

tests/unit_tests/steps/test_compile_c.py
MatthewHambley: Like anything related to mocking and patching, it has to be used with care, but it can be incredibly useful.

MatthewHambley:
> OK, I am stopping here and pass it back to you (extra long weekend) - main reason: could you explain to me the advantage of using `fs`?

The thing which `fs` brings to the party over `tmp_path` is that it is a memory filing system so it is much faster. Unit tests should, after all, be fast.

As a secondary advantage, being a whole filing system, you can use absolute filenames in safety which means you avoid having to add `tmp_path` to the front of everything. Maybe a small win.

I take your point about debugging and access to the generated files but I think the run-time advantages out way them since tests are run much more often than they are written.

MatthewHambley:
Note that GitHub C.I. has become a little confused as the Python versions to test with has changed. This leaves old deprecated versions in the list of required runs even though they aren't. I can force commit once you give an approval.

yaswant:
> Note that GitHub C.I. has become a little confused as the Python versions to test with has changed. This leaves old deprecated versions in the list of required runs even though they aren't. I can force commit once you give an approval.

@MatthewHambley I've removed 3.7 and 3.8 builds from the branch protection rules for develop. Time to include 3.11, 3.12 (or even 3.13.3 if needed) in the GitHub Action workflow #405

source/fab/dep_tree.py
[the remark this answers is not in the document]
hiker: Ah, thanks for the explanation.

source/fab/tools/tool_repository.py
```diff
@@ -75,7 +75,7 @@ def __init__(self):
         # Add the common shells. While Fab itself does not need this,
         # it is a very convenient tool for user configuration (e.g. to
         # query nc-config etc)
-        for shell_name in ["sh", "bash", "ksh", "dash"]:
+        for shell_name in ["sh", "bash", "ksh", "dash", 'zsh']:
```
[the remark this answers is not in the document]
hiker: I've just fixed up the PR that adds shell in the first place, and it just adds a single shell now. I'd expect we get a conflict when we merge this together later, and we can then fix this up to just list `sh`, as you suggested.

tests/unit_tests/steps/test_compile_c.py
hiker: I admit, I didn't like it initially, too - but got used to it.

source/fab/artefacts.py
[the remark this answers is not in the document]
hiker: Thanks for the explanation, and I agree with you :)

source/fab/tools/compiler_wrapper.py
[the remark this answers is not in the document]
hiker: Now I am confused :) I originally intended to find out what the bug exactly is that you mentioned, and check if this is still an issue with my latest compiler-profile branch ... but then it turns out that your version is now using `super.get_flags()` - which is what I wanted anyway. So, since this works here, all is good :)

source/fab/tools/compiler_wrapper.py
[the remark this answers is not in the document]
hiker: As per previous comment :)

source/fab/tools/compiler_wrapper.py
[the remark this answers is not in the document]
hiker: OK, it looks like you later merged in develop again, which then put the code I expected in place. So, I am not confused aymore, and otherwise happy :)

tests/unit_tests/steps/test_archive_objects.py
hiker: Thanks for the update !

hiker (approved):
Thanks for addressing my comments. Nice update. Proceeding to merge.

hiker:
Note that I have added this PR to the list of PRs to go into main (see https://github.com/MetOffice/fab/discussions/353). I think it would be best if we merge this in after the other 6 PRs that are currently waiting for review. It seems not to clash with the profiler profiles (surprisingly :) ). I think it's best if you don't delete the branch (and just updated it from main as the other 6 PRs go in), that should hopefully minimise conflicts.

PR #392, opened 2025-03-04
Update Python versions for testing

MatthewHambley:
We were still testing with 3.7 and 3.8 which have gone end-of-line. We were also not testing with modern supported versions.

Also prevent duplicate jobs running for both pull request and push.


allynt (approved):
LGTM

PR #395, opened 2025-03-06
Unify display of compiler wrapper

hiker:
This PR is mostly a two liner (remove `__str__ `  for compiler_wrapper, plus changing the corresponding tests) - so that compiler and compiler wrapper now have the name following the same template.

But this PR shows one additional change that I did in the 'add shell tool' PR earlier: Matthew didn't like that we supported `bash`, `ksh` etc, and recommended to only support `sh` as tool (that was when he reviewed this branch when going into develop). So since this `add_shell_tool` branch is now ready to be re-reviewed by him to get merged into main, I have removed the support for other shells (I checked the shell tools we use ... `tau_f90.sh`, `nf-config` etc, they are all `sh` only). 

The other branches (i.e. the ones already merged into develop, but not main) have already all been updated - but this is the next PR to be merged into develop, so it shows this change.


source/fab/tools/psyclone.py
```diff
@@ -149,7 +149,7 @@ def process(self,
                 # New version: no API, parameter, but -o for output name:
                 parameters.extend(["-o", transformed_file])
             else:
-                # 2.5.0 or earlier: needs api nemo, output name is -opsy
+                # 2.5.0 or earlier: needs api nemo, output name is -oalg
```
jasonjunweilyu, line 152: typo: `-opsy` instead of `-oalg`.

jasonjunweilyu (changes requested):
Hi @hiker , only one small typo and then this PR is good to be merged into `develop`.

jasonjunweilyu (approved):
Thanks @hiker . This is ready to be merged into `develop`.

Issue #398, opened 2025-03-14
Support llvm


hiker:
There is a bug in the definition of ifx:
```
@@ -456,7 +459,7 @@ class Ifx(FortranCompiler):
                          module_folder_flag="-module",
                          openmp_flag="-qopenmp",
                          syntax_only_flag="-syntax-only",
-                         version_regex=r"ifx \(IFORT\) (\d[\d\.]+\d) ")
+                         version_regex=r"ifx \(IFX\) (\d[\d\.]+\d) ")
```

icx is fine.

hiker:
Fixed in #401.

PR #400, opened 2025-03-14
Small bugfixes ifx worspace getter

### LINKED ISSUE #399: Add a method to get the fab working space
While there is already a function in util.py, this only tests `$FAB_WORKSPACE` or the default, we need to have a way of getting the actually used fab workspace from the build configuration.

There is already an attribute `project_workspace`, I would prefer to have this a proper getter.

hiker:
Fixes #399 and #398.


hiker:
Wrong branch

PR #408, opened 2025-05-05
Very new and fully clean shell tool

hiker:
This is a re-work of the original add_shell_tool_clean branch (which had incorrect merges applied and contained too much ).

Note that Matthew already reviewed this while it was being merged into develop, and as reaction to his feedback I have removed support for other shells (`bash`, ...), only leaving `sh`.


mo-nikosbaltas:
Checked the code and the changes implemented. It seems fine. Also unit test has been included.

PR #409, opened 2025-05-05
Standalone update to fparser 0.2

hiker:
Updates to fparser 0.2, and:
1. Removes code to check for statement sin OpenMP sentinels, since fparser now does this
2. Removes fparser "stop concatentation"  work around, which the new fparser doesn't need anymore.

This is a merge of the PRs update_to_fparser_0_2 and remove_fparser_concatenation.

Otherwise minor code cleanup.


cameronbateman-mo (approved):
A lot of changes but all seem sensible and related, good work.

cameronbateman-mo (approved):
Removed `iter_content` from `x90.py` file, was causing no attribute error.

hiker:
Hi @cameronbateman-mo , thanks a lot. The incorrect line in `x90.py` is confusing - it's properly done in develop, and the actual patch https://github.com/MetOffice/fab/commit/0b1ed17a
certainly removed that import. Guess git got confused somewhere?
Anyway, thanks for spotting and fixing that!!

PR #411, opened 2025-05-05
Use the same output format for wrapper as for normal compilers/tools.

hiker:
Tiny change that makes a compiler wrapper's `__str__ ` look the same as a compiler.


ericaneininger (approved):
No issues here.

PR #412, opened 2025-05-05
Standalone fix contained subroutine

hiker:
Bugfix, also replaces some fparser tree traversal functions with the ones provided by fparser.


Pierre-siddall (approved):
As the code concerns a refactor of how the nodes are processed when it comes to parsing Fortran files and refactors areas where the nodes where iterated over to use fparsers walk function instead of the previously defined iter_content function. Then the code sets up some unit tests to test these changes in an appropriate manner.  Therefore this looks good to be merged.

PR #413, opened 2025-05-05
Small ifx fix and add workspace getter

hiker:
Adds a convenience getter (to avoid actually addressing an attribute of an object), and bug fix for ifx.


MatthewHambley (approved):
Very minor changes, nothing controversial here.

PR #427, opened 2025-06-27
426 fix parallel testing of archive

### LINKED ISSUE #426: Executung unit_tests/steps/test_archive_objects.py in parallel causes errors
The above test cannot be exeuted in parallel. Reason is that it assumes that the `ar` tool has been tested if it is available. This happens if the tests are executed in serial in that file (since the very first tests will check availability, and then changes the flag in the instance).

But when executing in parallel, later tests might get executed before the first test checks for availability.

This patch fixes this by resetting the `ar` tool before each tests, and then testing that the `ar --version` call is done first.

hiker:
Fixes #426 (by resetting `ar` to be untested, and then checking that the `--version` command is executed first).

This is based on compiler_with_path


tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -82,20 +91,24 @@ def test_for_library(self, stub_tool_box,
...
         """
         Test that an incorrect archive tool is detected.
         """
         config = BuildConfig('proj', stub_tool_box)
-        cc = stub_tool_box.get_tool(Category.C_COMPILER, config.mpi, config.openmp)
-        # And set its category to be AR
-        cc._category = Category.AR
+        cc = stub_tool_box.get_tool(Category.C_COMPILER, config.mpi,
+                                    config.openmp)
+        # And set its category to be AR. Use monkeypatch since the compiler
```
Pierre-siddall, line 108: Could it be worth adding the link to the documentation here to differentiate between general monkey patching and pytest monkey patching, as well as helping developers understand what it involves ? 

```suggestion
        # And set its category to be AR. Use monkeypatch (https://docs.pytest.org/en/6.2.x/monkeypatch.html) since the compiler
```

Pierre-siddall (changes requested):
Thanks @hiker this looks like almost ready to be merged to me as the code looks great and has put in relevant guards to ensure that errors are not raised by previous tests which use the Ar tool. I only have one trivial suggestion to make it easier for developers down the line to understand what a monkeypatch is. Could you add a comment which takes you to the documentation for a pytest monkeypatch please?

Pierre-siddall:
Hi @hiker, before you add the small docs change could you please update your branch to be up to date with main to make sure all the CI checks still pass.

hiker:
All updated, thanks!

Pierre-siddall (approved):
Thanks for adding this @hiker , this looks good to be merged to me.

PR #430, opened 2025-06-30
422 module path search order

### LINKED ISSUE #422: Incorrect order of module path and compilation flags
When jules is configured without netcdf, it provides its own version of netcdf, i.e. a dummy module that implements the required functions, which just abort (or do nothing). So it has its own version of netcdf.mod.

- [ ]  The module output path is added to the end of the parameters, so any -I can result in using the wrong netcdf mod file
- [ ] Additionaly, fab adds first `$FFLAGS`, so a potential include path to a spack environment comes first, meaning the wrong .mod file is then used, even if -J is inserted at the front of the parameters.
- [ ] At least for gfortran, while -J is added to the module search path, it is apparently added to the END :cry: So we also need to add a `-I` as a first parameter in order to get the order right.

The following patch allowed me to compile jules without netcdf by addressing these three points
```
--- a/source/fab/tools/compiler.py
+++ b/source/fab/tools/compiler.py
@@ -118,6 +118,7 @@ class Compiler(CompilerSuiteTool):
 
         :returns: the flags to be used with this tool.'''
         fflags = os.getenv("FFLAGS", "").split()
+        return self._flags[profile]
         return self._flags[profile] + fflags
 
     def get_all_commandline_options(
@@ -413,8 +414,10 @@ class FortranCompiler(Compiler):
 
         # Append module output path
         if self._module_folder_flag and self._module_output_path:
-            params.append(self._module_folder_flag)
-            params.append(self._module_output_path)
+            params.insert(0, self._module_folder_flag)
+            params.insert(1, self._module_output_path)
+            params.insert(0, "-I")
+            params.insert(1, self._module_output_path)
 
         return params
```

Removal of environment variables (#332) should be done consistently first.

hiker:
This fixes #422:

We must have the fab-specific module include path first in the command line options (to support e.g. Jules providing its own dummy NetCDF modules - the dummy .mod file MUST be found before a system NetCDF mod file). Additionally, it looks like gfortran searches the module output path always last, so additionally we added an explicit include path for the Fab folder.

This allows Jules to compile with non-NetCDF support.

This is based on #428, so is blocked for now


hiker:
@yaswant , @t00sa ,@MatthewHambley  - ready for review, please assign

Issue #424, opened 2025-06-18
Updated command line tool


t00sa:
Create a new `fab` command line utility to replace the existing one.  Feature list includes:

- [ ] Recipe location
  - [ ] Allow `--file` to override default script location
  - [ ] Try to use `FabFile` in the current working directory if `--file` has not be set
  - [ ] Use a built-in zero-config recipe if there is no `FabFile`
- [ ] Improved logging
  - [ ] Create separate build and system loggers
  - [ ] Allow build and system levels to be set independently
  - [ ] Add `--quiet` mode to suppress almost all output
  - [ ] Add `--detailed` mode to show commands being executed
  - [ ] Log everything to a file in the project directory
- [ ] Improved progress indicators with the optional [rich](https://github.com/Textualize/rich) library
  - [ ] Add spinners to tasks of unknown duration
  - [ ] Add progress bars to multiprocessing pools
  - [ ] Use rich log handlers
  - [ ] Use rich generalised exception handling
- [ ] Extra command line options
  - [ ] Support build targets, assuming they are defined in the build recipe
  - [ ] Support `-DFOO=BAR` style command line options
  - [ ] Unrecognised options should be passed through to the build recipe
  - [ ] Support for getting help from from the build recipe argument parser, if any
- [ ] Basic recipe classes including:
  - [ ] A base class which sets up infrastructure for other recipes
  - [ ] A zero config recipe derived from the base class
  - [ ] LFRic-Baf support

t00sa:
**Note:** the branch currently contains some `sleep()` calls to slow things down, to allow the progress monitoring to be tested.  These will need to be removed before the changes go for review.

hiker:
Hi Sam, thanks a lot for this. I would really suggest that we should have a chat first, because a lot of the things you have done in Fab are already in BAF. For example, a zero-config setup based on BAF can be done with (for time reasons I didn't bother adding command line options to specify the name, I hard-coded "my_prog"):

```
class FabCli(BafBase):

    def __init__(self):
        super().__init__("my_prog")

    def setup_site_specific_location(self):
        '''This will likely not be necessary once the base class is in Fab, since it will know where to find the compiler-specific setup files'''
        root_dir = Path(__file__).parent
        sys.path.insert(0, str(root_dir))
        sys.path.insert(0, str(root_dir / "site_specific"))

    def find_source_files_step(self):      # That's the only real code required
        find_source_files(self.config,
                          source_root=".")

if __name__ == "__main__":
[...]

PR #450, opened 2025-07-23
Add build target classes to Fab

t00sa:
This adds a module directory containing build target classes.  The base module contains an abstract build implementation which is intended to be subclassed by the application to create a set of rules which can be used to compile an application.

It also creates a zero-config class derived from the base class which is intended to show how the rules should be implemented.

**Note:** This is independent of the other changes in #424

### DIFF
--- a/source/fab/target/zero.py
+++ b/source/fab/target/zero.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+
+"""
+Class which implements a zero-configuration build target.
+"""
+
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+
+from .base import FabTargetBase
+from ..tools import Category, ToolBox, ToolRepository
+from ..build_config import BuildConfig
+from ..steps.grab.folder import grab_folder
+from ..steps.find_source_files import find_source_files
+from ..steps.preprocess import preprocess_fortran
+from ..steps.c_pragma_injector import c_pragma_injector
+from ..steps.analyse import analyse
+from ..steps.compile_fortran import compile_fortran
+from ..steps.compile_c import compile_c
+from ..steps.link import link_exe
+
+
+class FabZeroConfig(FabTargetBase):
+    """Implementation of a zero-configuration build target."""
+
+    project_name = "zero-config"
+
+    @staticmethod
+    def add_arguments(parser: ArgumentParser) -> None:
+        """Add zero-config mode command line options.
+
+        Zero configuration requires the name of the directory
+        containing the source code.
+
+        :param ArgumentParser parser: an existing command line argument parser.
+        """
+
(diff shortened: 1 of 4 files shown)


source/fab/target/zero.py
```diff
@@ -0,0 +1,92 @@
...
+        tool_box.add_tool(fc)
+        tool_box.add_tool(linker)
+
+        with BuildConfig(
+            project_label=project_label,
+            mpi=False,
+            openmp=False,
+            tool_box=tool_box,
+            fab_workspace=Path(args.workspace),
+        ) as config:
```
hiker, line 82: Unfortunately, I was on leave and can only review now after this has been merged. Not sure if anyone sees this, but could someone please explain to me why this code duplicates what the base class introduces? For your reference:
```
     with self._config as _:
            self.grab_files_step()
            self.find_source_files_step()
            # This is a Fab function, which the user won't need to be
            # able to overwrite.
            c_pragma_injector(self.config)
            self.define_preprocessor_flags_step()
            self.preprocess_c_step()
            self.preprocess_fortran_step()
            self.analyse_step()
            self.compile_c_step()
            self.compile_fortran_step()
            # Disable archiving due to
            # https://github.com/MetOffice/fab/issues/310
            self.link_step()
```
Except for the base class introducing proper methods (which can be overwritten to add additional steps), that's basically identical :(
Looks like unnecessary technical debt here.

source/fab/target/zero.py
```diff
@@ -0,0 +1,92 @@
...
+
+        :param ArgumentParser parser: an existing command line argument parser.
+        """
+
+        parser.add_argument(
+            "source", type=Path, nargs="?", default=".", help="source directory"
+        )
+
+    @staticmethod
+    def check_arguments(parser: ArgumentParser, args: Namespace) -> None:
```
hiker, line 48: I don't really see how this class does a zero-config build. It contains three methods, that are not called. The check argument is static, so how can a derived class overwrite this to add its own validity tests?
I am probably missing something, but this PR feels incomplete. It adds some files without actually using it (short of the testing). Passing the parser in as argument doesn't feel very OO. Shouldn't the class have a parser? Then a derived class can add its own arguments etc.

tests/unit_tests/test_target.py
```diff
@@ -0,0 +1,107 @@
...
+
+    class FabTargetArgs(FabTargetBase):
+        """Test class with add_arguments."""
+
+        @staticmethod
+        def add_arguments(p):
+            """Add a positional argument."""
+            p.add_argument("test", type=str)
+
+    parser = argparse.ArgumentParser()
```
hiker, line 70: It feel very strange that the application has to create and provide its own `ArgumentParser`. Actually, this whole test is a bit odd: you basically call a static function in `FabTargetArgs`. Does deriving from `FabTargetBase` actually make any difference, or is any code of `FabTargetBase` actually used in this test? To me it looks like it is only testing `argparse`'s `add_argument` method, and nothing else. Am I missing something?

hiker (commented):
Obviously too late to do anything about it, but this PR feels incomplete, and incurring technical debt by duplicating code.

Issue #338, opened 2024-10-11
Working directory should default to current directory


MatthewHambley:
At the moment the working directory is, by default, placed in the root of the user's home directory. This is not obvious behaviour, particularly when operating in zero configuration mode.

A more obvious place to put it is in the current working directory.

hiker:
@MatthewHambley @yaswant @t00sa 

Given that this is a rather trivial change, but with huge impact on script development (tbh, a script should always set the location, but who knows :) ), should we add this to the next release?

PR #453, opened 2025-07-28
 Working directory should default to current directory

Pierre-siddall:
At the moment the fab workspace is created in the root of the users home directory. A more suitable place for the fab workspace to default to would be the current working directory. This pull request aims to implement a fix to this as outlined by #338.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,7 +69,7 @@ instance/
 .scrapy
 
 # PyBuilder
-target/
+# target/
 
 # Jupyter Notebook
 .ipynb_checkpoints
--- a/Documentation/source/advanced_config.rst
+++ b/Documentation/source/advanced_config.rst
@@ -11,29 +11,9 @@ aspects of software building may require more esoteric Fab features.
 
 Understanding the Environment
 =============================
-
-Fab uses well-known environment variables to identify tools and configure them.
-
-
-.. list-table:: Environment variables
-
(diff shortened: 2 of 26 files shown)


source/fab/util.py
jasonjunweilyu, line 1: Would it be better to keep the logger information about fab_workspace, at least for the debug level?

jasonjunweilyu (changes requested):
Hi @Pierre-siddall , thanks for modifying the `fab_workspace`. I only have one comment about possibly keeping the logging information, at least for debug. After this, then it is fine to me for this to go into `main`. As I don't have permission to merge to `main`, we may need to ask Yash or other people to do the merge at that point.

source/fab/util.py
Pierre-siddall: Hi @jasonjunweilyu , when @t00sa and I discussed this we though that the logging message didn't add much extra information to the output which couldn't already be inferred from the line above and therefore we took the view that it would be good to reduce the amount of potential clutter in the output.

jasonjunweilyu: I see. Thanks. Then I am fine with this. I thought that the logging message may be useful for first-time users to know how to set the fab workspace and why it defaults to the current directory. 
Just noticed that in Line 256, the description of `def get_fab_workspace()` would need to be updated. Then I think I can approve the PR.

PR #472, opened 2025-08-13
Fix output_fpath being cast to string in archive_objects

Pierre-siddall:
During https://github.com/MetOffice/fab/pull/431 implementation, it was found that output_fpath is always passed to Ar.create as a string, despite being type hinted as a Path. Currently, this requires creating a new Path instance from output_fpath to unlink archive (.a) files when Ar.create is called . Replacing the casting with a direct Path instantiation in archive_objects.py (lines 128-129) would be ideal, but since it involves usage of a Template instance, we need to verify that this change doesn't affect program semantics.

Once this PR is merged it shall close https://github.com/MetOffice/fab/issues/441 and will also close https://github.com/MetOffice/fab/issues/359.

This PR replaces #459 which I accidentally closed.


source/fab/artefacts.py
```diff
@@ -66,7 +66,7 @@ def reset(self):
                 self[artefact] = set()
 
     def add(self, collection: Union[str, ArtefactSet],
-            files: Union[Path, str, Iterable[Path], Iterable[str]]):
+            files: Union[Path, Iterable[Path]]):
```
hiker, line 69: I am not sure about this change. Yes, atm I believe we have only set of file paths in there (except for OBJECT_FILES/OBJECT_ARCHIVES, which is a dictionary), but since the same class might be used by user scripts, and we don't really know if a user only wants to store something else in there?

I am not opposed to this ;) But I think we should better discuss this change. Also, if it is done, we need to check the documentation, and at the very least add this restriction for the artefact set.

source/fab/artefacts.py
```diff
@@ -75,14 +75,15 @@ def add(self, collection: Union[str, ArtefactSet],
...
         elif not isinstance(files, Iterable):
-            # We need to use a list, otherwise each character is added
-            files = set([files])
+            files = {Path(files)}
 
         self[collection].update(files)
 
     def update_dict(self, collection: Union[str, ArtefactSet],
-                    values: Union[str, Iterable],
+                    values: Union[Path, Iterable[Path]],
```
hiker, line 86: Good catch with the Iterable typing :)

source/fab/steps/archive_objects.py
```diff
@@ -11,6 +11,7 @@
 import logging
 from string import Template
 from typing import Optional
+from pathlib import Path
```
hiker, line 14: I would recommend to sort the imports alphabetically within the same group of imports (group as in: system, ..., local projects). 
I don't think it's part of Fab's coding style, so I'll let this pass ... but if the modules are sorted, it's easier to see if you are importing a module twice.

source/fab/steps/archive_objects.py
```diff
@@ -121,12 +121,12 @@ def archive_objects(config: BuildConfig,
 
         if root:
             # we're building an object archive for an executable
-            output_fpath = str(config.build_output / f'{root}.a')
+            output_fpath = Path(str(config.build_output / f'{root}.a'))
```
hiker, line 124: You don't need `Path(str(...))` - `build_output` should be a Path, so using this with `/` and a string gives a path.

tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -60,7 +60,7 @@ def test_for_exes(self, stub_tool_box,
 
         # ensure the correct artefacts were created
         assert config.artefact_store[ArtefactSet.OBJECT_ARCHIVES] == {
-            target: {str(config.build_output / f'{target}.a')}
+            target: {Path(str(config.build_output / f'{target}.a'))}
```
hiker, line 63: As above, remove `Path(str())`.

tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -77,26 +77,24 @@ def test_for_library(self, stub_tool_box,
 
         help_command = ['ar', '--version']
         fake_process.register(help_command, stdout='1.0.0')
-        ar_command = ['ar', 'cr', '/fab/proj/build_output/mylib.a',
-                      'util1.o', 'util2.o']
+        ar_command = ['ar', 'cr', '/fab/proj/build_output/mylib.a', 'util1.o', 'util2.o']
```
hiker, line 80: Again, I didn't check with Fab's coding style, but general recommendation is 80 (https://peps.python.org/pep-0008/#maximum-line-length ... actually, 79) characters, which I think was the reason why this was split where it was.

tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -77,26 +77,24 @@ def test_for_library(self, stub_tool_box,
...
-            ArtefactSet.OBJECT_FILES, {'util1.o', 'util2.o'}, None
+            ArtefactSet.OBJECT_FILES, {Path('util1.o'), Path('util2.o')}, None
         )
 
         with warns(UserWarning,
                    match="_metric_send_conn not set, cannot send metrics"):
             archive_objects(config=config,
-                            output_fpath=config.build_output / 'mylib.a')
-
+                            output_fpath=Path(config.build_output / 'mylib.a'))
```
hiker, line 92: No `Path()` required.

tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -77,26 +77,24 @@ def test_for_library(self, stub_tool_box,
...
         assert call_list(fake_process) == [help_command, ar_command]
 
         # ensure the correct artefacts were created
         assert config.artefact_store[ArtefactSet.OBJECT_ARCHIVES] == {
-            None: {str(config.build_output / 'mylib.a')}}
+            None: {Path(str(config.build_output / 'mylib.a'))}}
```
hiker, line 97: No `Path(str(...))`

tests/unit_tests/steps/test_archive_objects.py
```diff
@@ -115,7 +113,7 @@ def test_incorrect_tool(self, stub_tool_box, monkeypatch):
 
         with raises(RuntimeError) as err:
             archive_objects(config=config,
-                            output_fpath=config.build_output / 'mylib.a')
+                            output_fpath=Path(config.build_output / 'mylib.a'))
```
hiker, line 116: No `Path(..)`.

hiker (changes requested):
Thanks a lot for starting to clear some of the typing problems up.

You extended the scope a bit (restricting Artefact Store to only have Path), which I think should either be discussed, but I am also happy to leave this to a second review from someone else.

Otherwise a few minor code cleanups

source/fab/artefacts.py
Pierre-siddall: Thanks @hiker for this I'll hold off tinkering with this for the time being and we can raise this at the next standup.

source/fab/steps/archive_objects.py
hiker: No, you are right that this should become a Path, not a `str`. Just check:
```
output_fpath = config.build_output / f'{root}.a'
reveal_type(output_fpath)
```
For me this is a `Path`.

hiker:
@Pierre-siddall , I am happy to progress this without the change to the ArtefactStore - just leave everything as `str` or `Path` (I believe it might need some additional typing updates :) ), but the rest is certainly a nice improvement. We can tackle additional typing issues later.

source/fab/steps/archive_objects.py
hiker: You missed this line, the `Path(str(...))` is not necessary, `config.build_output / f'{root}.a'` should be a Path already.

source/fab/steps/archive_objects.py
hiker: I am happy to ignore this, given that sorting the imports alphabetically (i.e. pathlib before typing) is not part of our coding standards :)

source/fab/artefacts.py
hiker: Dang, we forgot to discuss this today. @MatthewHambley , do you have an opinion if restricting the ArtefactStore to only store Paths is ok? I admit I can't think of a good example, but wanted your input before this change is being done.

hiker (changes requested):
You missed one line with an unnecessary casting. Also, since we forgot to discuss this in today's meeting, I would at least prefer to get @MatthewHambley opinion on the restriction that Artefacts must be Paths. I am ok with it, but I am just wondering if there might be use cases where a string would be required.

source/fab/artefacts.py
MatthewHambley: I think artefacts are always files. If it turns out they aren't that probably implies a bigger problem needing a more extensive refactor.

MatthewHambley (dismissed):
With changes noted by @hiker .

hiker (changes requested):
One change still missing as far as I can see.

Pierre-siddall:
This should be good to go now, apologies about misinterpreting the casting issue for `Path(str(...))`  I didn't release the `build_output` handled the concatenation of strings.

hiker (approved):
Thanks a lot for this, I like clearing up these typing issues. I am also quite happy with the 'side effect' of turning the Artefacts to store only Paths, it certainly helps with making sure that it is used as intended.

PR #476, opened 2025-08-14
Add compiler selection arguments

t00sa:
Add some standard flags for compiler selection, with environment variables as potential fallbacks.

Adds tests for the new options and add tests for two previously untested error conditions to bring test coverage up to 100 percent.


MatthewHambley (changes requested):
The primary issue is the default compilers.

t00sa:
I'll check to see if this is still needed and close the PR if it isn't...

t00sa:

PR #478, opened 2025-08-15
Add ability to cache command line arguments

t00sa:
This adds a feature which allows complex command line arguments to be written to a cache file and then re-used.  Where an argument is specified in both the cache and on the current command line, the command line version takes priority.

Cache arguments themselves are not sticky, so if --save-cache and --use-cache are set on the command line, they will get written to the cache but ignored when the cache is next loaded.  This prevents the cache from being inadvertently overwritten with new arguments.

### DIFF
--- a/source/fab/cui/arguments.py
+++ b/source/fab/cui/arguments.py
@@ -111,7 +111,7 @@ class _FabArgumentCache:
             # Do not add the group if caching is disabled
             return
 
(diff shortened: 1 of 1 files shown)


t00sa:
Thanks @hiker, I'm glad you think this is useful.  I'm going to pick this up again shortly, integrate your suggestions, and get the PR moving again.

t00sa:

PR #480, opened 2025-08-19
379 select linker

hiker:


tests/unit_tests/tools/test_tool_repository.py
```diff
@@ -221,7 +223,12 @@ def test_default_gcc_suite(category, fake_process: FakeProcess) -> None:
 
     tr = ToolRepository()
     tr.set_default_compiler_suite('gnu')
-    def_tool = tr.get_default(category, mpi=False, openmp=False)
+    if category == Category.LINKER:
+        def_tool = tr.get_default(category, mpi=False, openmp=False,
+                                  enforce_fortran_linker=True)
+    else:
+        def_tool = tr.get_default(category, mpi=False, openmp=False)
```
t00sa, line 230: I'm unclear why the linker is separated out.  The test should be the same if you allow `enforce_fortran_linker` to use its default value.

tests/unit_tests/tools/test_tool_repository.py
```diff
@@ -238,7 +245,12 @@ def test_default_intel_suite(category, fake_process: FakeProcess) -> None:
 
     tr = ToolRepository()
     tr.set_default_compiler_suite('intel-classic')
-    def_tool = tr.get_default(category, mpi=False, openmp=False)
+    if category == Category.LINKER:
+        def_tool = tr.get_default(category, mpi=False, openmp=False,
+                                  enforce_fortran_linker=True)
+    else:
+        def_tool = tr.get_default(category, mpi=False, openmp=False)
```
t00sa, line 252: As per the previous comment, I'm unclear why the linker is separated out.  The test should be the same if you allow `enforce_fortran_linker` to use its default value.

tests/unit_tests/tools/test_tool_repository.py
```diff
@@ -284,3 +296,17 @@ def test_tool_repository_full_path(fake_process: FakeProcess) -> None:
...
+    '''
+    tr = ToolRepository()
+    gfortran = tr.get_tool(Category.FORTRAN_COMPILER, "/usr/bin/gfortran")
+    assert isinstance(gfortran, Gfortran)
+    assert gfortran.name == "gfortran"
+    assert gfortran.exec_name == "gfortran"
+    assert gfortran.exec_path == Path("/usr/bin/gfortran")
+
+    fake_process.register(['/usr/bin/gfortran', 'a'])
+    gfortran.run("a")
```
t00sa, line 312: Does this require an assert to check the correctness of the command?

t00sa (commented):
Looks promising.  Please see the review comments inline.

tests/unit_tests/steps/test_link.py
```diff
@@ -11,51 +11,145 @@
...
+                          symbol_defs=["foo"])
+    config.artefact_store[ArtefactSet.BUILD_TREES] = \
+        {"foo": {a_f.fpath: a_f}}
+
+    config.artefact_store[ArtefactSet.OBJECT_FILES] = \
+        {'foo': {'foo.o', 'bar.o'}}
+
+    with warns(UserWarning,
+               match="_metric_send_conn not set, cannot send metrics"):
+        link_exe(config, libs=['mylib'], flags=['-fooflag', '-barflag'])
```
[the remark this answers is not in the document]
hiker: Strictly speaking not, in line 74 we register the expected call. If the call would not have the right parameters, we get an error. But I totally agree that this is just too 'hidden', so I've added an explicit assert at the end.

tests/unit_tests/steps/test_link.py
```diff
@@ -11,51 +11,145 @@
...
+                    symbol_defs=["main"])
+    config.artefact_store[ArtefactSet.BUILD_TREES] = \
+        {"foo": {a_c.fpath: a_c}}
+
+    config.artefact_store[ArtefactSet.OBJECT_FILES] = \
+        {'foo': {'foo.o', 'bar.o'}}
+
+    with warns(UserWarning,
+               match="_metric_send_conn not set, cannot send metrics"):
+        link_exe(config, libs=['mylib'], flags=['-fooflag', '-barflag'])
```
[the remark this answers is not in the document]
hiker: Same reason, same fix :)

tests/unit_tests/tools/test_tool_repository.py
hiker: Same reason, same fix :)

source/fab/steps/link.py
```diff
@@ -65,6 +65,10 @@ def link_exe(config,
     source_getter = source or DefaultLinkerSource()
     target_objects = source_getter(config.artefact_store)
 
+    if len(target_objects) == 0:
+        logger.warning("No target objects defined, linking aborted")
```
t00sa, line 69: If this is a fatal error, would it be better as an exception?

source/fab/steps/link.py
```diff
@@ -87,8 +91,10 @@ def link_exe(config,
     logger.info(f'Linker is {linker.name}')
 
     libs = libs or []
-    if flags:
-        linker.add_post_lib_flags(flags, config.profile)
+    flags = flags or []
+    source_getter = source or DefaultLinkerSource()
+
+    target_objects = source_getter(config.artefact_store)
```
t00sa, line 97: This appears to be a duplicate of the code at lines 65 and 66

tests/unit_tests/fab_base/test_fab_base.py
```diff
@@ -278,7 +279,9 @@ def test_workspace(monkeypatch, change_into_tmpdir) -> None:
     # succeed.
     # Note that the project directories are only created once
     # build is called.
-    fab_base.build()
+    with caplog.at_level(logging.WARNING):
+        fab_base.build()
+    assert "No target objects defined, linking aborted" in caplog.text
```
t00sa, line 284: See comment about raising an exception rather than a log message

tests/unit_tests/fab_base/test_fab_base.py
```diff
@@ -358,7 +361,9 @@ def test_build_binary(monkeypatch) -> None:
         patcher = mock.patch(f"fab.fab_base.fab_base.{function_name}")
         mocks[function_name] = (patcher, patcher.start())
 
-    fab_base.build()
+    with caplog.at_level(logging.WARNING):
+        fab_base.build()
+    assert "No target objects defined, linking aborted" in caplog.text
```
t00sa, line 366: See comment about raising an exception rather than a log message

tests/unit_tests/steps/test_link.py
```diff
@@ -153,3 +156,35 @@ def test_run_select_linker_c(fake_process: FakeProcess,
...
+    tool_box.add_tool(linker)
+
+    config = BuildConfig('link_test', tool_box, fab_workspace=Path('/fab'),
+                         mpi=False, openmp=False, multiprocessing=False)
+
+    with warns(UserWarning,
+               match="_metric_send_conn not set, cannot send metrics"):
+        with caplog.at_level(logging.WARNING):
+            link_exe(config, libs=['mylib'], flags=['-fooflag', '-barflag'])
+    assert "No target objects defined, linking aborted" in caplog.text
```
t00sa, line 188: See comment about raising an exception rather than a log message

t00sa (commented):
I've added some comments about the log warning versus raising an exception.  If the code cannot be linked either because there are no files, then it may be better to force an obvious failure at this point.

I also realise that this is probably part of a wider discussion about exceptions in fab, so I'm happy to accept logging of a message in this change and we can revisit it in a later ticket.

source/fab/steps/link.py
hiker: Agreed. Since your new exceptions aren't merged, I've used a ValueError (?? seems to be nest, the value of the target symbols are wrong)

source/fab/steps/link.py
hiker: Oops - done.

tests/unit_tests/fab_base/test_fab_base.py
hiker: Change to exception.

hiker: Fixed.

Issue #479, opened 2025-08-19
"DEPENDS ON:" c file dependency can not be ignored during the analyse step


jasonjunweilyu:
Currently, the `ignore_mod_deps` parameter in `@step def analyse` only affects Use statements for ignoring third-party Fortran modules. 

Now `shumlib` has become a dependency that comes from the build environment. When building `lfric_inputs` with FAB, `c_shum_byteswap` that should be ignored during dependency analysis, according to https://code.metoffice.gov.uk/trac/lfric_apps/browser/main/trunk/applications/lfricinputs/Makefile#L30, can not be ignored by FAB with the error: 
```
Traceback (most recent call last):
  File "/scratch/hc46/hc46_gitlab/lfric_fab/lfric_source_53477/source/apps/applications/lfricinputs/./fab_lfric2um.py", line 24, in <module>
    fab_lfric_inputs.build()
  File "/scratch/hc46/hc46_gitlab/builds/tBfW9y6q/0/bom/ngm/lfric/lfric_atm-fab/lfric-baf/fab/source/fab/fab_base.py", line 703, in build
    self.analyse_step()
  File "/scratch/hc46/hc46_gitlab/lfric_fab/lfric_source_53477/source/apps/applications/lfricinputs/fab_lfricinputs.py", line 186, in analyse_step
    super().analyse_step(ignore_mod_deps=lfric_inputs_ignore_mod_deps)
  File "/scratch/hc46/hc46_gitlab/lfric_fab/lfric_source_53477/source/core/infrastructure/build/fab/lfric_base.py", line 373, in analyse_step
    analyse(self.config, root_symbol=self.root_symbol,
  File "/scratch/hc46/hc46_gitlab/builds/tBfW9y6q/0/bom/ngm/lfric/lfric_atm-fab/lfric-baf/fab/source/fab/steps/__init__.py", line 26, in wrapper
    func(*args, **kwargs)
  File "/scratch/hc46/hc46_gitlab/builds/tBfW9y6q/0/bom/ngm/lfric/lfric_atm-fab/lfric-baf/fab/source/fab/steps/analyse.py", line 174, in analyse
    add_mo_commented_file_deps(project_source_tree)
  File "/scratch/hc46/hc46_gitlab/builds/tBfW9y6q/0/bom/ngm/lfric/lfric_atm-fab/lfric-baf/fab/source/fab/mo.py", line 37, in add_mo_commented_file_deps
    f.file_deps.add(lookup[dep].fpath)
[...]

PR #481, opened 2025-08-20
479 added check of ignore_mod_deps for "DEPENDS ON:" comment analysis

jasonjunweilyu:
This PR is for the issue #479.

This PR has added the check of ignore_mod_deps for "DEPENDS ON:" comment analysis. It has also added a test for the ignore_mod_deps parameter of Fortran Analyser, which confirms that fortran modules in USE statements and "DEPENDS ON:"  and c file in "DEPENDS ON:"  can be ignored.

### DIFF
--- a/source/fab/parse/fortran.py
+++ b/source/fab/parse/fortran.py
@@ -195,7 +195,9 @@ class FortranAnalyser(FortranAnalyserBase):
         :param std:
             The Fortran standard.
         :param ignore_mod_deps:
-            Module names to ignore in use statements.
+            Module names to ignore in use statements or
+            'DEPENDS ON' files to ignore or 'DEPENDS ON'
+            modules to ignore.
 
         """
         super().__init__(config=config,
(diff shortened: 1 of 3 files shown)


hiker (dismissed):
Thanks for this contribution, it is an important bug fix since (since lfric_input cannot be compiled with Fab anymore with this PR - cause is the change to use shumlib as a library and not compiling it as part of the code anymore).

I would prefer that the parameter `ignore_mod_deps` is renamed to (e.g.) `ignore_dependencies`, since it now includes module names and file names. Given that Jason is on leave for around two weeks, I have taken the liberty to add this to the PR. I'll approve the PR, but it must now be reviewed from someone else.

I would actually also prefer to change the coding style (80 character line length), but the files involved here have a huge number of these (>80 lines in one case), so it would take a small change to a file to a rather gigantic change. I'll leave it to the reviewer to decide if we should follow the "you touch it, you fix" it policy, I am happy to fix this up.

PR #484, opened 2025-08-21
Detect and error on duplicate module definitions

Pierre-siddall:
Instead of warnings being used when duplicate modules are detected, ideally the analysis code in analyse.py would give an error forcing the build to be aborted (as module names must be unique in order for dependency analysis to work).

This PR aims to fix this issue and once merged will close https://github.com/MetOffice/fab/issues/406.

### DIFF
--- a/source/fab/steps/analyse.py
+++ b/source/fab/steps/analyse.py
@@ -292,20 +292,22 @@ def _add_manual_results(special_measure_analysis_results, analysed_files: Set[An
         logger.info(f'added {len(special_measure_analysis_results)} manual analysis results')
 
 
-def _gen_symbol_table(analysed_files: Iterable[AnalysedDependent]) -> Dict[str, Path]:
+def _gen_symbol_table(analysed_files: Iterable[AnalysedDependent], debug=False) -> Dict[str, Path]:
     """
     Create a dictionary mapping symbol names to the files in which they appear.
 
     """
     symbols: Dict[str, Path] = {}
-    duplicates = []
+    duplicates = False
     for analysed_file in analysed_files:
         for symbol_def in analysed_file.symbol_defs:
             # check for duplicates
             if symbol_def in symbols:
-                duplicates.append(ValueError(
-                    f"duplicate symbol '{symbol_def}' defined in {analysed_file.fpath} "
-                    f"already found in {symbols[symbol_def]}"))
+                if debug:
+                    logger.warning(
+                        f"duplicate symbol '{symbol_def}' defined in {analysed_file.fpath} "
+                        f"also found in {symbols[symbol_def]}\n")
+                duplicates = True
                 continue
             symbols[symbol_def] = analysed_file.fpath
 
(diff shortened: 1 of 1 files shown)


source/fab/steps/analyse.py
```diff
@@ -292,29 +292,31 @@ def _add_manual_results(special_measure_analysis_results, analysed_files: Set[An
         logger.info(f'added {len(special_measure_analysis_results)} manual analysis results')
 
 
-def _gen_symbol_table(analysed_files: Iterable[AnalysedDependent]) -> Dict[str, Path]:
+def _gen_symbol_table(analysed_files: Iterable[AnalysedDependent], debug=False) -> Dict[str, Path]:
```
MatthewHambley, line 295: I don't understand the value of the `debug` argument. It seems to provide optional warnings when duplicates found. Much better would be to use the logging framework we already have to log the warnings. Then the user decides whether they see them in the usual manner.

tests/unit_tests/steps/test_analyse.py
```diff
@@ -44,15 +44,8 @@ def test_duplicate_symbol(self,
...
-        with warns(UserWarning):
-            result = _gen_symbol_table(analysed_files=analysed_files)
-
-        assert result == {
-            'foo_1': Path('foo.c'),
-            'foo_2': Path('foo.c'),
-            'bar_1': Path('bar.c'),
-            'bar_2': Path('bar.c'),
-        }
+        with raises(RuntimeError):
```
MatthewHambley, line 47: It is common all over the Fab code but `RuntimeError` is something we want to get away from. It's too generic. In this case `ValueError` may be more appropriate but we have `FabException` (`from fab import FabException`) available.

MatthewHambley (changes requested):
Some issues around error handling.

source/fab/steps/analyse.py
```diff
@@ -70,7 +70,6 @@ def analyse(
         std: str = "f2008",
         special_measure_analysis_results: Optional[Iterable[FortranParserWorkaround]] = None,
         unreferenced_deps: Optional[Iterable[str]] = None,
-        ignore_dependencies: Optional[Iterable[str]] = None,
```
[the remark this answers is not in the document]
Pierre-siddall: Not sure why I removed that while trying to remove `ignore_mod_deps` will get this fixed now.

source/fab/steps/analyse.py
```diff
@@ -109,9 +109,6 @@ def analyse(
         For example, functions that are called in a one-line if statement.
         Assuming the files containing these symbols are present and analysed,
         those files and all their dependencies will be added to the build tree(s).
-    :param ignore_dependencies:
-        Third party Fortran module names in USE statements, 'DEPENDS ON' files
-        and modules to be ignored.
```
MatthewHambley, line 114: The argument was restored but the comment is still removed.

source/fab/steps/analyse.py
```diff
@@ -132,8 +129,7 @@ def analyse(
 
     # todo: these seem more like functions
     fortran_analyser = FortranAnalyser(config=config,
-                                       std=std,
-                                       ignore_dependencies=ignore_dependencies)
+                                       std=std)
```
MatthewHambley, line 132: Since the `ignore_dependencies` argument was restored the use of it should probably be restored too.

tests/unit_tests/steps/test_analyse.py
```diff
@@ -44,15 +44,8 @@ def test_duplicate_symbol(self,
...
 
-        with warns(UserWarning):
-            result = _gen_symbol_table(analysed_files=analysed_files)
-
-        assert result == {
-            'foo_1': Path('foo.c'),
-            'foo_2': Path('foo.c'),
-            'bar_1': Path('bar.c'),
-            'bar_2': Path('bar.c'),
-        }
```
MatthewHambley, line 55: Is this information preserved in the exception? If so it would be good to test for it.

MatthewHambley (changes requested):
Incomplete change from previous review and "one last thing."

PR #486, opened 2025-08-22
Replace RuntimeErrors with custom exceptions

t00sa:
Swap all RuntimeError exceptions for custom fab exceptions using a class hierarchy derived from RuntimeError to minimise disruptive changes.

These changes fall into three broad areas:

* creation of a new set of error classes in `fab.errors` and an associated set of tests
* changes to the existing regression tests to:
  * check the new message formats
  * check the exceptions - which are still caught as `RuntimeError` to test the impact on calling functions - match the expected instance of the new error class
* replacement of `raise RuntimeError` across the code base with the new error classes

While the change is complete and the test suite runs successfully in my environment, I still have a couple of outstanding questions that need to be addressed in review:

* are the names of the new exceptions classes acceptable?  The top level class names end with `Error` but the derived classes do not.  Ending every class name with `Error` would make fab more consistent with python, at the cost of longer exception names
* are there any potential clashes or issues with the in-flight baf PR?

### DIFF
--- a/source/fab/errors.py
+++ b/source/fab/errors.py
@@ -61,14 +61,14 @@ class FabToolMismatch(FabToolError):
 
     def __init__(
         self,
-        tool: Union["Category", "Tool", str],
-        category: Union["Category", "Tool", type, str],
+        toolname: str,
+        category: Union["Category", "Tool", type],
         expected: str,
     ) -> None:
         self.category = category
         self.expected = expected
 
-        super().__init__(tool, f"got type {category} instead of {expected}")
+        super().__init__(toolname, f"got type {category} instead of {expected}")
 
 
 class FabToolInvalidVersion(FabToolError):
@@ -85,7 +85,7 @@ class FabToolInvalidVersion(FabToolError):
 
     def __init__(
         self,
-        tool: Union["Category", "Tool", str],
+        toolname: str,
         value: str,
         expected: Optional[str] = None,
     ) -> None:
@@ -96,7 +96,7 @@ class FabToolInvalidVersion(FabToolError):
         if expected is not None:
             message += f" should be {repr(expected)}"
 
-        super().__init__(tool, message)
+        super().__init__(toolname, message)
 
 
 class FabToolPsycloneAPI(FabToolError):
@@ -167,7 +167,7 @@ class FabToolInvalidSetting(FabToolError):
     def __init__(
         self,
         setting_type: str,
-        tool: Union["Category", "Tool", str],
+        category: Union["Category", str],
         additional: Optional[str] = None,
     ) -> None:
         self.setting_type = setting_type
@@ -176,7 +176,7 @@ class FabToolInvalidSetting(FabToolError):
         if additional:
             message += f" {additional}"
 
-        super().__init__(tool, message)
+        super().__init__(category, message)
 
 
 class FabUnknownLibraryError(FabError):
@@ -209,16 +209,13 @@ class FabCommandError(FabError):
 
     def __init__(
         self,
-        command: str,
+        command: List[str],
         code: int,
         output: Union[str, bytes, None],
         error: Union[str, bytes, None],
-        cwd: Optional[Union[str, Path]] = None,
+        cwd: Optional[Union[Path, str]] = None,
     ) -> None:
-        if isinstance(command, list):
-            self.command: str = " ".join(command)
-        else:
-            self.command = str(command)
+        self.command: str = " ".join(command)
         self.code = int(code)
         self.output = self._decode(output)
         self.error = self._decode(error)
@@ -245,7 +242,7 @@ class FabCommandNotFound(FabError):
         preserved for inspection by the caller.
     """
 
-    def __init__(self, command: str) -> None:
+    def __init__(self, command: Union[List[str], str]) -> None:
         self.command = command
         if isinstance(command, list):
             self.target: str = command[0]
@@ -271,9 +268,7 @@ class FabMultiCommandError(FabError):
     :param label: an identifier for the multiprocessing section
     """
 
-    def __init__(
-        self, errors: List[Union[str, Exception]], label: Optional[str] = None
-    ) -> None:
+    def __init__(self, errors: List[Exception], label: Optional[str] = None) -> None:
         self.errors = errors
         self.label = label or "during multiprocessing"
 
@@ -314,11 +309,13 @@ class FabSourceMergeError(FabSourceError):
         targeted
     """
 
-    def __init__(self, tool: str, reason: str, revision: Optional[str] = None) -> None:
+    def __init__(
+        self, tool: "Tool", reason: str, revision: Optional[str] = None
+    ) -> None:
         self.tool = tool
         self.reason = reason
 
-        message = f"[{tool}] merge "
+        message = f"[{tool.name}] merge "
         if revision:
             message += f"of {repr(revision)} "
         message += f"failed: {reason}"
--- a/tests/unit_tests/test_errors.py
+++ b/tests/unit_tests/test_errors.py
@@ -140,9 +140,6 @@ class TestCommandErrors:
         err = FabCommandError(["ls", "-l", "/nosuch"], 1, b"", b"ls: cannot", "/")
         assert str(err) == "return code 1 from 'ls -l /nosuch'"
 
-        err = FabCommandError("ls -l /nosuch", 1, None, "ls: cannot", "/")
-        assert str(err) == "return code 1 from 'ls -l /nosuch'"
-
     def test_not_found(self):
         """Test command not found errors."""
 
@@ -185,10 +182,13 @@ class TestSourceErrors:
     def test_merge(self):
         """Test merge errors."""
 
-        err = FabSourceMergeError("git", "conflicting source files")
+        tool = Mock()
+        type(tool).name = PropertyMock(return_value="git")
+
+        err = FabSourceMergeError(tool, "conflicting source files")
         assert str(err) == "[git] merge failed: conflicting source files"
 
-        err = FabSourceMergeError("git", "conflicting source files", "vn1.1")
+        err = FabSourceMergeError(tool, "conflicting source files", "vn1.1")
         assert str(err) == "[git] merge of 'vn1.1' failed: conflicting source files"
 
     def test_fetch(self):
--- a/source/fab/steps/grab/svn.py
+++ b/source/fab/steps/grab/svn.py
@@ -125,6 +125,6 @@ def check_conflict(tool: Versioning, dst: Union[str, Path]):
             for element in entry:
                 if (element.tag == 'wc-status' and
                         element.attrib['item'] == 'conflicted'):
-                    raise FabSourceMergeError("svn", xml_str)
+                    raise FabSourceMergeError(tool, xml_str)
 
     return False
--- a/source/fab/tools/versioning.py
+++ b/source/fab/tools/versioning.py
@@ -109,7 +109,7 @@ class Git(Versioning):
             self.run(['merge', 'FETCH_HEAD'], cwd=dst, capture_output=False)
         except RuntimeError as err:
             self.run(['merge', '--abort'], cwd=dst, capture_output=False)
-            raise FabSourceMergeError("git", str(err), revision) from err
+            raise FabSourceMergeError(self, str(err), revision) from err
 
 
 # =============================================================================


MatthewHambley:
I'm about to start a review but I'll leave this comment here for immediate consideration. Are you aware of `FabException` in https://github.com/MetOffice/fab/blob/main/source/fab/__init__.py? Have you considered how this interacts with your change?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+    :param tool: name of the current tool
+    :param category: name of the current category
+    :param expectedd: name of the correct category
+    """
+
+    def __init__(
+        self,
+        tool: Union["Category", "Tool", str],
+        category: Union["Category", "Tool", type, str],
```
MatthewHambley, line 67: Why can argument `Tool` have a value of type `Category` and argument `category` have value of type `Tool`? Also, why allow strings?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+        expected: str,
```
MatthewHambley, line 68: Isn't this an expected `Category`?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+    :param tool: name of the current tool
+    :param value: output from the query command
+    :param expected: optional format of version string
+    """
+
+    def __init__(
+        self,
+        tool: Union["Category", "Tool", str],
+        value: str,
+        expected: Optional[str] = None,
```
MatthewHambley, line 92: Should `tool` have type `Tool`? is `expected` a regular expression?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+            message += f" should be {repr(expected)}"
+
+        super().__init__(tool, message)
+
+
+class FabToolPsycloneAPI(FabToolError):
+    """PSyclone API and target problem.
+
+    Error when the specified PSyclone API, which can be empty to
+    indicate DSL mode, and the associated target do not match.
```
MatthewHambley, line 108: I know this is how PSyclone works but we might prefer to be explicit about the API.

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+    Error where a tool which is not available in a particular suite of
+    tools has been requested.
+
+    :param tool: name of the current tool
+    :param suite: optional name of the current tool suite
+    """
+
+    def __init__(
+        self, tool: Union["Category", "Tool", str], suite: Optional[str] = None
```
MatthewHambley, line 148: `tool` of type `Tool`?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+    Error where an invalid setting, e.g. MPI, has been requested for a
+    particular tool.
+
+    :param setting_type: name of the invalid setting
+    :param tool: the tool to which setting applies
+    :param additional: optional additional information
+    """
+
+    # FIXME: improve these here and in tool_repository
```
MatthewHambley, line 167: Ideally we will eventually remove the need for this failure mode in that the API will not allow impossible requests to be made. Clearly not an issue for this change.

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+    def __init__(
+        self,
+        setting_type: str,
+        tool: Union["Category", "Tool", str],
```
MatthewHambley, line 172: `tool` has type `Tool`?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+    :param command: the command being run
+    :param code: the command return code from the OS
+    :param output: output stream from the command
+    :param error: error stream from the command
+    """
+
+    def __init__(
+        self,
+        command: str,
```
MatthewHambley, line 214: The subprocess module also allows commands to be specified as a list of strings. Check how it is done when this exception is raised - we might prefer to accept it straight rather than having to convert it.

In fact, looking below, I see you cover this eventuality so reflect it in the type hinting.

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+        code: int,
+        output: Union[str, bytes, None],
+        error: Union[str, bytes, None],
+        cwd: Optional[Union[str, Path]] = None,
+    ) -> None:
+        if isinstance(command, list):
+            self.command: str = " ".join(command)
+        else:
+            self.command = str(command)
+        self.code = int(code)
```
MatthewHambley, line 224: Both these things are being cast to the type they are already. Is there a reason for that?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+        self.output = self._decode(output)
+        self.error = self._decode(error)
+        self.cwd = cwd
```
MatthewHambley, line 227: Two types are accepted but we should store just one. I suggest `Path`.

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+    Error where the target command passed to a subprocess call could
+    not be found.
+
+    :param command: the target command.  For clarity, only the first
+        item is used in the error message but the entire command is
+        preserved for inspection by the caller.
+    """
+
+    def __init__(self, command: str) -> None:
+        self.command = command
```
MatthewHambley, line 251: Homogenise argument to either string or list of strings. Latter would remove the need for the test to extract target.

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+class FabMultiCommandError(FabError):
+    """Unpack multiple exceptions into a single one.
+
+    Error which combines all potential exceptions raised by a
+    multiprocessing section into a single exception class for
+    subsequent inspection.
+
+    This feature is is required because versions of python prior to
```
MatthewHambley, line 269:
```suggestion
    This feature is required because versions of python prior to
```

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+    3.11 do not support ExceptionGroups.
```
MatthewHambley, line 270: Maybe mark this as `Todo:` so someone can be prompted to remove it once 3.10 is no longer supported.

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+
+    :param errors: a list ot exceptions
+    :param label: an identifier for the multiprocessing section
+    """
+
+    def __init__(
+        self, errors: List[Union[str, Exception]], label: Optional[str] = None
```
MatthewHambley, line 277: If this exception consolidates exceptions, why is it accepting strings?

If that is useful, should it create an exception to hold the string so `self.errors` holds only exceptions?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+    conflict that requires manual resolution.
+
+    :param tool: name of the version control system
+    :param reason: reason/error output from the version control system
+        indicating why the merge failed
+    :param revision: optional name of the specific revision being
+        targeted
+    """
+
+    def __init__(self, tool: str, reason: str, revision: Optional[str] = None) -> None:
```
MatthewHambley, line 319: `tool` has type `Tool`?

source/fab/steps/grab/prebuild.py
```diff
@@ -28,4 +29,4 @@ def grab_pre_build(config, path, allow_fail=False):
         msg = f"could not grab pre-build '{path}':\n{err}"
         logger.warning(msg)
```
MatthewHambley, line 30: Do we want to log a warning *and* raise an exception?

tests/unit_tests/test_cui_arguments.py
```diff
@@ -101,7 +101,7 @@ def test_nonexistent_user_file(self, fs: FakeFilesystem, capsys):
         assert exc.value.code == 2
 
         captured = capsys.readouterr()
-        assert "error: fab file does not exist" in captured.err
+        assert "fab file does not exist" in captured.err
```
MatthewHambley, line 104: Where we are testing error messages I think we should test as much as possible. i.e. prefer equality over "in".

tests/unit_tests/test_errors.py
```diff
@@ -0,0 +1,179 @@
...
+
+        err = FabToolError("cc", "compiler message")
+        assert str(err) == "[cc] compiler message"
+
+        # Mock defines an internal name property, so special handling
+        # is required to reset the value to support testing
+        tool = Mock(_name="tool cc")
+        del tool.name
+        err = FabToolError(tool, "compiler message")
+        assert str(err) == "[tool cc] compiler message"
```
MatthewHambley, line 61: This seems alarming. Is this related to Joerg's comment about peeking at private state?

tests/unit_tests/test_errors.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class TestCommandErrors:
+    """Test various command errors."""
+
+    def test_command(self):
+        """Test FabCommandError in various configurations."""
+
+        err = FabCommandError(["ls", "-l", "/nosuch"], 1, b"", b"ls: cannot", "/")
+        assert str(err) == "command 'ls -l /nosuch' returned 1"
```
MatthewHambley, line 122: Would it be easier to read if the sentence was restructured to put the variable length element last. e.g. "Return code 1 from command: ls -l /nosuch".

This way the eye knows where to go to for the error code and the start of the command.

tests/unit_tests/test_errors.py
```diff
@@ -0,0 +1,179 @@
...
+        assert str(err) == "no source files found after filtering"
+
+    def test_merge(self):
+        """Test merge errors."""
+
+        err = FabSourceMergeError("git", "conflicting source files")
+        assert str(err) == "[git] merge failed: conflicting source files"
+
+        err = FabSourceMergeError("git", "conflicting source files", "vn1.1")
+        assert str(err) == "[git] merge of vn1.1 failed: conflicting source files"
```
MatthewHambley, line 173: Maybe quote marks around revision?

tests/unit_tests/tools/test_linker.py
```diff
@@ -124,7 +124,7 @@ def test_get_lib_flags_unknown(stub_c_compiler: CCompiler) -> None:
     linker = Linker(compiler=stub_c_compiler)
     with raises(RuntimeError) as err:
         linker.get_lib_flags("unknown")
-    assert str(err.value) == "Unknown library name: 'unknown'"
+    assert str(err.value) == "unknown library unknown"
```
MatthewHambley, line 127: This might benefit from a redrafting. Initial upper-case and maybe quote marks around library name.

MatthewHambley (changes requested):
This is an extensive change which tidies up exception handling considerable. A few questions and suggestions included in-line.

In general, we have generated API documentation but is there anything which could be added (or removed) from the static documentation in light of this change?

source/fab/errors.py
```diff
@@ -0,0 +1,344 @@
...
+Custom exception classes designed to replace generic RuntimeError
+exceptions originally used in fab.
+"""
+
+from pathlib import Path
+from typing import List, Optional, TYPE_CHECKING, Union
+
+if TYPE_CHECKING:
+    from fab.tools.category import Category
+    from fab.tools.tool import Tool
```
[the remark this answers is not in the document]
t00sa: It's certainly unpleasant.  I _think_ it's what I meant when I mentioned the dependency later on.  I think Joerg has plans to deal with it at some point?

source/fab/errors.py
t00sa: I _think_ this is because the `RuntimeError` for which this is a drop-in replacement uses a mixture of tools, categories, and strings in a not particularly intuitive way.  I don't have a strong opinion on whether to leave as is or whether to implement some additional error classes for each pattern.

source/fab/errors.py
t00sa: I agree: it's not particularly elegant, but I wasn't sure whether it was worth adding more plumbing to raise different sorts of exceptions vs having a generic exception that handles everything

t00sa (commented):
Clearly, some things I still need to tidy up.  I'll get on with them tomorrow and update the PR when they're ready.

.github/workflows/build.yml
```diff
@@ -71,6 +71,12 @@ jobs:
           pip install .[dev]
           pip install .[docs]
 
+      # Some of the system tests require git to be configured
+      - name: Set up git configuration
+        run: |
+          git config --global user.email "fab"
+          git config --global user.name "Fab Testing"
```
yaswant: I see you need this for `git merge`. There is no potential issue here, but does this have to be --global config? I know there is no strict validation here, but wondering if fab@github.action as user.email would be better?

source/fab/errors.py
t00sa: Tool is also always provided as a string, so I'll update the type hinting accordingly.

t00sa:
I have addressed the comments from the previous reviews, the CI tests pass, and I've successfully used my branch to build skeleton in LFRic core using Jason's branch, so I think this ready for another round of reviewing.

.github/workflows/build.yml
t00sa: It probably doesn't have to be, but this is what git suggests when the configuration items don't exist

hiker:
Removed 'ready for review' label, since it's still a draft.

t00sa:

PR #488, opened 2025-08-27
Refactor documentation build process and update copyright year

### LINKED ISSUE #487: Deploy documentation copies wrong pdf
In #470 we create PDF of the documentation and link it to the main HTML. However, the workflow does not copy the correct PDF due to a flaw in logic. We now convert scalable graphics to PDF and this creates additional PDFs (e.g., `analysis_results_hierarchy.svg -> analysis_results_hierarchy.pdf` and this the files that gets copied across.

Should be easy to fix.

yaswant:
Fixes #487

### DIFF
--- a/Documentation/source/conf.py
+++ b/Documentation/source/conf.py
@@ -84,7 +84,7 @@ html_theme_options = {
     "footer_start": ["crown-copyright"],
     "footer_center": [
         "sphinx-version",
-        f"Updated: {datetime.datetime.now().strftime('%Y-%m-%d %H:%M')}",
+        "last-updated",
     ],
     "footer_end": ["theme-version"],
 }
--- a/Documentation/source/conf.py
+++ b/Documentation/source/conf.py
@@ -12,6 +12,7 @@
 #
 import os
 import sys
+import datetime
 
 from fab import __version__ as fab_version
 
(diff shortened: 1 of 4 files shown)


Documentation/source/conf.py
```diff
@@ -81,7 +82,10 @@
         }
     ],
     "footer_start": ["crown-copyright"],
-    "footer_center": ["sphinx-version"],
+    "footer_center": [
+        "sphinx-version",
+        f"Updated: {datetime.datetime.now().strftime('%Y-%m-%d %H:%M')}",
```
t00sa, line 87: I don't think this is correct: the list entries should reference templates file names rather than literal strings.  I think you need to add:

```python
html_last_updated_fmt = '%Y-%m-%d %H:%M'
```

Before this section and then change these lines to:

```suggestion
        "sphinx-version",
        "last-updated",
```

t00sa (commented):
The make process fails with an exception when I run it interactively:

```
sphinx.errors.ExtensionError: Handler <function update_and_remove_templates at 0x7fbac7871940> for event 'html-page-context' threw an exception (exception: 'Updated: 2025-09-01 21:06.html' not found in ...
```

I've suggested a fix in the comments.

t00sa (approved):
Looks good.  Approved.

PR #490, opened 2025-08-29
Move to using `which` to determin tool presence

MatthewHambley:
PSyclone and compilers still manage version numbers as before as they are interesting to us. This should also address BoM's concern regarding broken Python installations.

Also notice that the `ToolRepository` causes a lot of trouble but fixing (or potentially just removing) that is a larger issue.


MatthewHambley:
This change has been overtaken by a mountain of technical debt. Other changes are needed before this one can be made and they will render this branch obsolete.

Issue #515, opened 2025-10-23
FabBase.analysis_step does not support ignore_dependencies


hiker:
LFRicAtm needs to pass through a list of dependency. ATM, it calls `analyse` directly (not `analyse_step` in the base class), but it would be good if the various `analyse_step` functions would accept identical parameters, and LFRicAtm could all the base `analyse_step` implementation instead.

PR #516, opened 2025-10-23
#515 Added optional ignore_dependencies parameter to analyse_step.

hiker:
A tiny change, only required to avoid `mypy` warnings in LFRic build scripts (and to allow the lfric_atm script to call `analyse_step` instead of the `analyse` function directly).

@yaswant , ready for review.


t00sa (approved):
Thanks!  Looks good.

PR #535, opened 2026-01-12
Remove insecure workflow

MatthewHambley:
This workflow was flagged as insecure by an automatic check. I'm not sure what it's for, some sort of container I think. So the simplest solution is to get rid of it.


t00sa:
Originally added three years ago by PR #206.  We don't use the singularity images, so I don't think there's a reason to keep this.

t00sa (dismissed):
Approving but waiting for feedback from @yaswant before merging.

yaswant (dismissed):
I've tried several approaches but haven't been able to get this action to work. I think it's fine to remove this workflow from the repository.

yaswant (approved):

Issue #311, opened 2024-06-17
Improve support for user-efined tools


hiker:
With #309 in progress, we need to implement a way to allow users to specify their own tools.

Suggested approach:
1. TBH, the fine-grained tool categories (fcm ,git, subversion etc each being their own cateforie) would appear a probably useless overkill. Instead, add all these tools as MISC.
2. Allow more than one MISC tool in the tool box.

With this implement, a user can add arbitrary tools (all as `MISC`). This works fine as long as the name of the tools are unique (i.e. we get the tool `svn` or `subversion` from the `MISC` category).

MatthewHambley:
If separate categories are overkill for none compiler tools the next obvious question is "Why have them for tools?" Is everything "misc"?

hiker:
In lfric_core, I need to add a category to add a pfunit tool. I opted to replace the implementation of Category with one that allows to extend categories. This way, any application can add their own tool and pass them around (allowing each site to e.g. replace the implementation of a tool with their own if required).

Required changes are only in Category, the API it provides remains the same.

PR #553, opened 2026-04-13
#311 Make Categories extendable

hiker:
This makes the `Category` 'enum' extendable, i.e. allows application to put its own tools into the tool repository. Original use case is adding a pFUnit tool (for lfric_core pFUnit support).

Alternatively, we could re-type all `category` arguments to be `Category|str`, but that feels very hacky to me. This implementation supports the same API as an enum (i.e. it's a bit more complicated than strictly necessary, but means no other code needs to be adjusted). I have successfully used this in an lfric_core branch.

### DIFF
--- a/source/fab/tools/category.py
+++ b/source/fab/tools/category.py
@@ -7,6 +7,8 @@
 '''This simple module defines an Enum for all allowed categories.
 '''
 
+from __future__ import annotations
+
 from typing import Optional
 
 
@@ -42,14 +44,16 @@ class Category(int, metaclass=CategoryMeta):
             # Called via __reduce__ (i.e. pickle), restore
             # the original int value
             obj = super().__new__(cls, val)
+        # New name. If it already exists, return the existing
+        # object
+        elif name in cls._values:
+            obj = cls._values[name]
         else:
-            # New name. Verify that it doesn't exist yet
-            if name in cls._values:
-                raise ValueError(f"Category '{name}' already exists.")
             # Get a new id for the name. Use +1 to avoid using a zero
             # (just in case)
             obj = super().__new__(cls, len(cls._values) + 1)
-        cls._values[name] = obj
+            cls._values[name] = obj
+
         return obj
 
     def __reduce__(self):
@@ -71,6 +75,15 @@ class Category(int, metaclass=CategoryMeta):
         # an attribute with the same name
         self._name = name
         setattr(Category, name, self)
+
+    @staticmethod
+    def add(name: str) -> None:
+        """
+        Adds a new category.
+        """
+        # We don't need to store the instance, it is added as an attribute
+        # to this class anyway.
+        Category(name)
 
     def __str__(self):
         return self._name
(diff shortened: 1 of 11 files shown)


hiker:
I had to make some more adjustments, since things failed when using subprocesses. I think (un)pickling caused the Category attributes to be different when running in subprocesses (i.e. while `category == Category.AR` might be true, `category is Category.AR` was not).

TBH, this class is now rather overcomplicated imho. I am more than happy to re-implement this with a simpler class, which will:
- require to replace one or two `category is ...` with `category ==...`
- replaces `category.is_compiler` with something like `Category.is_compiler(category)`

These two usages which worked fine with enum causes a lot of the complications in the new implemenation.

@t00sa , @MatthewHambley - quick feedback welcome, so that I can either document the new behaviour, or re-implement using a simpler class.

source/fab/tools/category.py
```diff
@@ -7,33 +7,120 @@
...
+    """
+
+    # A dictionary used for iterating over all enums.
+    _values: dict[str, "Category"] = {}
+
+    def __iter__(cls):
+        return iter(cls._values.values())
+
+
+class Category(int, metaclass=CategoryMeta):
```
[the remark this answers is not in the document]
hiker: Hmm - I just took that out, since the Category is now an `int`, and as such this is used for hash and eq, which is sufficient (since the ints are uniquely mapped to the categories), and I thought it would be easier to remove these two methods. But I am equally happy to add them back in.

source/fab/tools/category.py
```diff
@@ -7,33 +7,120 @@
...
+    def __new__(cls, name: str, val: Optional[int] = None):
+        # choose a numeric value for the int part
+        if val is not None:
+            # Called via __reduce__ (i.e. pickle), restore
+            # the original int value
+            obj = super().__new__(cls, val)
+        else:
+            # New name. Verify that it doesn't exist yet
+            if name in cls._values:
+                raise ValueError(f"Category '{name}' already exists.")
```
MatthewHambley: This might be a `KeyError`.

Thinking about it a bit further; is defining a category which already exist actually an error. The user's goal in defining a category is that the category exists after they've done it. It doesn't seem to matter whether it exists before they do it. As long as the definition doesn't clear all the tools already registered to it.

source/fab/tools/category.py
```diff
@@ -7,33 +7,120 @@
...
+    FCM: "Category"
+    FORTRAN_COMPILER: "Category"
+    FORTRAN_PREPROCESSOR: "Category"
+    GIT: "Category"
+    LINKER: "Category"
+    MISC: "Category"
+    PSYCLONE: "Category"
+    RSYNC: "Category"
+    SHELL: "Category"
+    SUBVERSION: "Category"
```
MatthewHambley: You may not need to quote the class name if you `from __future__ import annotations`.

source/fab/tools/category.py
```diff
@@ -7,33 +7,120 @@
...
+Category("FCM")
+Category("FORTRAN_COMPILER")
+Category("FORTRAN_PREPROCESSOR")
+Category("GIT")
+Category("LINKER")
+Category("MISC")
+Category("PSYCLONE")
+Category("RSYNC")
+Category("SHELL")
+Category("SUBVERSION")
```
MatthewHambley: I'm not sure about this paradigm of adding a category by creating and throwing away an object. Is there a reason for not just providing a class method.

Also, since categories are now being generated on the fly it would make sense, particularly for those categories which will only ever have one entry (like `RSYNC`) to define them in the module relating to that tool.

MatthewHambley (changes requested):
Some thoughts about design and implementation.

hiker:
@MatthewHambley , thanks a lot for the review. I had a look at all Fab scripts that have been merged or are WIP atm (lfric_core, lfric_apps, jules, um, socrates) - none of them uses an 'is' test or `is_compiler`. As such, I will re-implement this functionality and update a few lines in Fab accordingly. I think _kiss_ should win here. I will take your comments into account in the new PR. I'll leave this open till the new PR is ready for now.

hiker:
OK, I am back-paddling :) When I tried a simpler implementation, in the end I ran into all the problems that this PR actually solved, so I gave up, and am just fixing up this PR following your comments.

source/fab/tools/category.py
hiker: No problem. This PR is now supporting re-creating an existing category, in which case the existing object will be returned. I added (modified the previous failure test) a test to make sure we do indeed get the same instance (which is required for `category is Category.XXX` tests.)

source/fab/tools/category.py
hiker: Oh yes, indeed. Fixed.

source/fab/tools/category.py
hiker: I have added a static method install, so this is now `Category.add("GIT")` etc.

I have also moved the single-item categories to be created in the corresponding classes. Then I have also removed the `MISC` category (which was intended for users to add new tools, which is now not required anymore since a user can create their own categories now). But the unit testing needed a test-category, so I have added a special category `CATEGORY_FOR_UNIT_TESTS`, which is added in the test that needs it.

hiker:
Ready for next review,

hiker:
Note that with the decision to move pfunit into Fab, this PR is not particularly urgent. It could be useful, but will not be required for pfunit support in lfric_core anymore (though of course the upcoming PR adding pfunit to Fab will be required instead ;) )

MatthewHambley (approved):
This looks like it streamlines the category system a bit.

PR #556, opened 2026-04-16
555 multiple c main programs

### LINKED ISSUE #555: Multiple C main programs
While Fab can handle multiple Fortran main program, it can't handle multiple C main programs (since a main program tries to add `main` as a symbol, so you end up with duplicated symbol). This affect the Socrates build script (which has many small utilities in C).

My suggested solution would be to add `main@filename.c` (or so) as symbol, instead of just `main`. This way Fab should be able to handle multiple main programs. It needs a thorough check though (analysis step, build artefacts, ...), linking - to see if this all will work as expected.

If anyone has any comments (or better suggestion), I appreciate any feedback before I start implementing this.

hiker:
Fixes #555. I've updated/added tests for multiple C programs, and can also confirm that this version is able to build multiple Socrates C executables.

I have also fixed the symbol handling (C symbols are case dependent, only Fortran symbols are not). And as a result of the change the output filename of a C main program is now changed from `main` to be the same as the main source file name (without `.c`).

### DIFF
--- a/Documentation/source/advanced_config.rst
+++ b/Documentation/source/advanced_config.rst
@@ -598,7 +598,7 @@ symbol and add it, *and all its dependencies*, to every :term:`Build Tree`.
     :linenos:
 
     ...
-    analyse(state, root_symbol='my_prog', unreferenced_deps=['my_func'])
+    analyse(state, root_symbols='my_prog', unreferenced_deps=['my_func'])
     ...
 
 Unparsable Files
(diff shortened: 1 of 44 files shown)


hiker:
I've labeled it for the UM release. While it is strictly speaking, it's Socrates that needs it, not UM (and socrates is already built as part of UM/lfric as required), but it's a small change only, so I don't think we need it as a separate mile stone.

allynt (approved):
Adds the ability to cope w/ multiple `main` C programs by appending the corresponding filename to each identifier in the symbol collection;  For example, "main@my_filename".

LGTM.

PR #558, opened 2026-04-23
557 contained subroutines are not external

hiker:
Prevents subroutines/functions that are 'contained' in another subroutine to be exported as defined external symbols. This is a requirement for Socrates, which has dozen of subroutines which each define their own `subroutine nf(...)`, which atm results in Fab complaining about duplicated symbols.

We already exclude subroutines in modules and interfaces, so it is just a small change to also exclude anything that's an `Internal_Subprogram_Part` (i.e. it is `contained` in another program unit).

I also did some cleanup of pylint errors in a test file.


MatthewHambley (approved):
As advertised, a straight forward change.

PR #561, opened 2026-04-24
559 support external declarations

hiker:
Adds support for external declarations.

### DIFF
--- a/Documentation/source/advanced_config.rst
+++ b/Documentation/source/advanced_config.rst
@@ -598,7 +598,7 @@ symbol and add it, *and all its dependencies*, to every :term:`Build Tree`.
     :linenos:
 
     ...
-    analyse(state, root_symbol='my_prog', unreferenced_deps=['my_func'])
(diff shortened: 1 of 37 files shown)


hiker:
This is blocked by #558 (it modifies the same file, so will likely cause conflicts otherwise)

hiker:
I've also added support for `external` attributes (which is from the Fortran grammar point of view different from an external statement), and fixed an issue that subroutines in interfaces where not recorded as dependencies - we clearly lacked a test case there, which I have added.

MatthewHambley (approved):
Adds to a previous change. Seems to make sense.

MatthewHambley (changes requested):
I notice that this branch is not at head of trunk. It's probably best if the developer performs the merge.

hiker:
> I notice that this branch is not at head of trunk. It's probably best if the developer performs the merge.

Yes, it was blocked by 558 (do you prefer me to mark them as 'draft' instead?). I've merged it all together, should be good to go now

MatthewHambley (approved):
A straight forward change.

Issue #552, opened 2026-04-13
Add pfunit tool to Fab


hiker:
I have developed a pFUnit tool for LFRic (to be added to lfric_core). I wonder if it would be useful to have that tool added to Fab, to make it easier (and consistent) to support pFUnit for non-lfric codes?

Note that this is only a tool to make the pFUnit API easily accessible, it's not adding automatic pFUnit building anywhere, that would still be up to the actual app.

hiker:
Merged a while ago.

PR #562, opened 2026-05-05
#552 Added pfunit tool and tests.

hiker:
As discussed, this adds the pfunit tool to Fab.

Note that I have NOT based this on #311 (PR #553). IMHO, the latter is kind of optional.  Once we have pFUnit, we don't need really need #311 (it's just a nice to have for the future ;) ).

I am setting this as draft for now, to make sure that the core-fab-with-pfunit branch works as expected.


hiker:
There is a certain "asymmetry": this class has a function for getting the include flags, but not the libraries. I would welcome some feedback here. Ideally, we should have a kind of 'dependency' manager, which handles linker AND compiler flags. So far, we only have library flags (in the linker). That would be a bit of work, time we all might not have atm.

The consistent solution would be to leave it to the site-specific setup to add the include flags require for pfunit, and I am happy to do that. In lfric, I have a mixin class which automatically adds the include path from pfunit, so it removes the need to add the include flags in all site-specific files (and if a site need something else, they can just add their own pfunit implementation). So, it basically means no need to change any site-specific setup files. But admittedly, it is inconsistent.

Not sure what's the best way forward here :)

But it is working with skeleton now, so marking this as ready for review.

source/fab/tools/category.py
```diff
@@ -26,6 +26,7 @@ class Category(Enum):
     AR = auto()
     RSYNC = auto()
     SHELL = auto()
+    PFUNIT = auto()
```
MatthewHambley: Now that we have your change to categories on trunk, do you wont to reconsider your position on using them here?

source/fab/tools/pfunit.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    This is a class to encapsulate pFUnit. It relies on the environment
+    variable $PFUNIT to indicate the location of the source code.
+    This is required since besides .mod files and executable, it also
+    contains the source code for a Fortran driver program .
+    It assumes that pFUnit's preprocessor `funitproc` is in $PFUNIT/bin.
+    """
+
+    def __init__(self):
+        pfunit_home = os.environ.get("PFUNIT", "")
```
MatthewHambley, line 33: In the future we might provide an argument on this constructor to specify the location of the driver source (that's what we need `PFUNIT` for). We could then move the use of the environment variable to the top level build script, or not. But I don't think it's worth making that change how.

MatthewHambley (changes requested):
A question and some food for thought.

source/fab/tools/pfunit.py
hiker: I am happy to wait (or if you have details what you want, I can add this in, doesn't sound too complicated).

source/fab/tools/category.py
hiker: Yes, I've brought this up-to-date with what is done for the other tools.

source/fab/tools/pfunit.py
```diff
@@ -0,0 +1,72 @@
+##############################################################################
+# (c) Crown copyright Met Office. All rights reserved.
+# For further details please refer to the file COPYRIGHT
+# which you should have received as part of this distribution
+##############################################################################
+
+"""This file contains the Rsync class for synchronising file trees.
```
MatthewHambley: In response to this documentation I say "no it doesn't." Looks like a copy-and-paste mistake.

MatthewHambley (changes requested):
One minor documentation niggle and I see static type checking and unit testing are failing.

source/fab/tools/pfunit.py
hiker: Oops :) Fixed.

hiker:
> One minor documentation niggle and I see static type checking and unit testing are failing.

Ah, no - the wrong/unexpected order of applying the PRs ;) There default `Tool` base class has no flags anymore (as in, the ability for an application to set flags), that's now in `ToolWithFlags`. Since pfunit doesn't need the application to provide special flags (it only offers `-quiet` as option), I didn't 'rebase' it to be a `ToolWithFlags`, instead I removed parts of the test that checked for flag handling. Would be easy to change if required.

MatthewHambley (approved):
This looks ready to go now.

PR #565, opened 2026-05-09
544 include LFRic_core and UM updates

hiker:
This adds changes and improvements done, based on the code review for lfric_core and UM (plus pfunit for lfric core).

It's mostly a collection of small changes, but I've renamed `root_symbol` to `root_symbols`, since this has always been a list, and is frequently used this way (e.g. um builds `um` and `recon`; socrates has over 50 binaries, ...). I left the old names in place with a deprecation warning for now.


t00sa:
Makes sense for me to review this, given I worked on both the UM and core tickets

t00sa (approved):
Changes look good and test suite passes.  Approved!

Issue #568, opened 2026-05-21
Re-introduce MISC category


hiker:
Some lfric_core tools rely on using the MISC category (i.e. tools like rose_picker, where only one version will ever exist, and which will never be added to a tool box). It feels better to re-introduce the MISC category and set it as default for a tool.

hiker:
Fixed as part of #497

PR #569, opened 2026-05-21
#568 Re-introduce MISC category, and make it default for a tool.

hiker:
This fixes current build failures in lfric_core (some tools rely on the default MISC category), and given that we don't have a new Fab release, it feels easier to support this in Fab than doing another lfric_core PR to define a category.


hiker:
Confirmed that lfric builds again with these changes.

hiker:
I am closing this PR, since I had to merge it into #497 (the pkg_config tool needs (and it makes sense) the MISC category), and it is a tiny change anyway.

PR #574, opened 2026-06-19
Pin actions

### LINKED ISSUE #573: Pin actions
Pinning actions in a GitHub Actions workflow by using a full 40-character Git commit SHA instead of a mutable tag (like `@v4`) or branch name (like `@main`) is the industry standard for securing software supply chains.

yaswant:
closes #573

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,6 @@ jobs:
     name: Code Quality
     permissions:
       contents: read # Requires read access to fetch source files
-      pull-requests: write # Required to post coverage metrics, test results, or status logs back into PR comments
 
     runs-on: ubuntu-24.04
     strategy:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,24 +4,26 @@ on:
   push:
     branches:
       - main
-      - v*.*
+      - 'v*.*'
   pull_request:
     branches:
       - main
-      - v*.*
+      - 'v*.*'
 
 concurrency:
   # cancel any in-progress runs for the same group (branch/tag) {{github.ref}}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -4,24 +4,26 @@ on:
...
+
 jobs:
   checks:
-
+    name: Code Quality
     permissions:
-      contents: read
-      pull-requests: write
+      contents: read # Requires read access to fetch source files
+      pull-requests: write # Required to post coverage metrics, test results, or status logs back into PR comments
```
MatthewHambley: I don't think we do any of these things in this repository. In which case, is it a good idea to provide the permission?

MatthewHambley (commented):
This seems to make sense.

.github/workflows/build.yml
yaswant: Well spotted @MatthewHambley 
In its current state, this is superfluous. Do we have plan to post a comment in PR e.g., if the coverage drops below a certain threshold? If not I can remove the permission now.

MatthewHambley: We would certainly consider it, but not immediately.

Ideally GitHub's own coverage reporting (which is on trial at the moment) would be extended to work on forked pull requests. Then we would use that.

MatthewHambley (approved):
This is a straight forward change.

.github/workflows/build.yml
yaswant: makes sense. write permission on pr now removed in https://github.com/MetOffice/fab/pull/574/commits/0e9e9fcd44c81673fcf260cd2b1c762effec7347

MatthewHambley (approved):
Happy with this.

PR #575, opened 2026-06-19
support python 3.14

### LINKED ISSUE #510: Support Python 3.14
Just adding version 3.14 to the testing matrix does not work. Some of our tests fail and it's not instantly obvious why. I suspect the fact that the "subprocess" module has changed default engine with this version may be the reason.

Rather than try to fix this while we drag around a mountain of technical debt it would be best to postpone it for a while. It still needs doing though.

hiker:
Closes #510

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,7 @@ from pytest import fixture
 from pytest_subprocess.fake_process import FakeProcess, ProcessRecorder
 
 from fab.build_config import BuildConfig
+from fab.tools.category import Category
 from fab.tools.compiler import CCompiler, FortranCompiler
 from fab.tools.linker import Linker
 from fab.tools.tool_box import ToolBox
@@ -194,6 +195,11 @@ def stub_tool_repository(stub_fortran_compiler,
     monkeypatch.setattr(stub_c_compiler, 'check_available', return_true)
     monkeypatch.setattr(stub_linker, 'check_available', return_true)
     tool_repository = ToolRepository()
+    # Remove all compiler and linker, since they might exist on the system
+    # running these tests, which can break tests.
+    for cat in [Category.C_COMPILER, Category.FORTRAN_COMPILER,
+                Category.LINKER]:
+        tool_repository[cat] = []
     tool_repository.add_tool(stub_fortran_compiler)
     tool_repository.add_tool(stub_c_compiler)
     tool_repository.add_tool(stub_linker)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
 
     steps:
       # This step does not need a GA cache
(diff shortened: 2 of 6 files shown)


tests/conftest.py
```diff
@@ -194,6 +195,11 @@ def stub_tool_repository(stub_fortran_compiler,
     monkeypatch.setattr(stub_c_compiler, 'check_available', return_true)
     monkeypatch.setattr(stub_linker, 'check_available', return_true)
     tool_repository = ToolRepository()
+    # Remove all compiler and linker, since they might exist on the system
+    # running these tests, which can break tests.
+    for cat in [Category.C_COMPILER, Category.FORTRAN_COMPILER,
+                Category.LINKER]:
+        tool_repository[cat] = []
```
MatthewHambley, line 202: How does this change relate to Python v3.14?

hiker:
@t00sa , I don't fully understand the cui/aeguments.py file, so please check that my change there was ok.

I also fixed a small test bug (some tests would fail if you have a working mpif90 in your path, see conftest.py)

MatthewHambley (commented):
In general I am not in favour of code under test being different from deployed code but I think that is a problem we can postpone to later.

tests/conftest.py
hiker: As I've written:
> I also fixed a small test bug (some tests would fail if you have a working mpif90 in your path, see conftest.py)

Not directly related, except my python 3.14 environment happened to have mpif90 and caused test failures.

hiker:
> In general I am not in favour of code under test being different from deployed code but I think that is a problem we can postpone to later.

I totally agree with that, I only did it because of the existing comment acknowledging the line, and so I thought it would be acceptable for now (the commit that introduced the change only mentions something about making it aware of ifort or so. nothing really helpful as far as I could see).

I actually did try a few other things, but no real success. The problem seems to be that we have two level of child processes (multiprocessing spawn for the steps, then we run compiler etc), and Copilot told me (and admittedly I didn't check, but this is pretty much what I see):

> pytest‑subprocess has never supported mocking subprocesses inside child processes.
Python 3.14’s stricter import/pickle rules make this limitation more visible.

E.g. we fake:
```
        fake_process.register(['sfc', '--version'], stdout='1.2.3')
        fake_process.register(['sfc', fake_process.any()])
```
and then call `compile_pass`, which does `run_mp` and then checks if the compiler is available (first fake process), and then the actual compilation. So even if we set the compiler to be available (avoiding the first fake process), we are still stuck with the second one :(

More googling might have given me a different approach:  I tried before to switch multiprocessing back to fork, but that somehow didn't work (I believe the complication is to find where multiprocessing is imported first). But what seems to work:
in `steps/__init__.py` (`run_mp`):
```
    if config.multiprocessing and not no_multiprocessing:
        ctx = multiprocessing.get_context("fork") 
        with ctx.Pool(config.n_procs) as p:
            results = p.map(func, items)
    else:
        results = [func(f) for f in items]

    return results
```
I.e. explicitly switching from the new forkserver back to fork. I don't know enough about the details of this change to know if this is acceptable, but it feels like a better solution - at least then tests and 'real' code are consistent. But using fork might have disadvantages (i.e. there must be a reason why the python default multiprocessing was switched).

Let me know what you prefer me to do (and if I should leave the line in the build config in that disables subprocessing in case of debugging)?

MatthewHambley (approved):
One thing at a time, lets get Python 3.14 support in place.

PR #578, opened 2026-06-22
Abort analyse step when parse errors occur

### LINKED ISSUE #577: Fab does not abort in case of analyse errors
I am not sure if this is done intentional, but Fab does not abort when an analysis error happens (see #576). This wastes potentially time (i.e. Fab might try to compile, even though one file was not analysed, and then a dependency is missing, and compilation error happens due to unknown symbols/missing .mod files).

A quick look indicates that the same would happen with the PSyclone step if I am not mistaken.

It should be reasonable easy to fix by adding a call to `check_error`, as e.g. is done in preprocessing.

jordansilly77-stack:
## Summary
- stop the analyse step when parser errors are returned
- reuse the existing `check_for_errors` helper for consistent step behaviour
- update the unit test to assert parse errors halt the build

Closes #577.

## Tests
- `.venv/bin/python -m pytest tests/unit_tests/steps/test_analyse.py tests/unit_tests/steps/test_steps.py`
- `.venv/bin/python -m flake8 source/fab/steps/analyse.py tests/unit_tests/steps/test_analyse.py`

I also ran `.venv/bin/python -m pytest tests/unit_tests/steps`; unrelated `test_find_source_files` cases fail locally because no Fortran compiler is available.

### DIFF
--- a/source/fab/steps/analyse.py
+++ b/source/fab/steps/analyse.py
@@ -46,7 +46,7 @@ from fab.mo import add_mo_commented_file_deps
 from fab.parse import AnalysedFile, EmptySourceFile
 from fab.parse.c import AnalysedC, CAnalyser
 from fab.parse.fortran import AnalysedFortran, FortranParserWorkaround, FortranAnalyser
-from fab.steps import run_mp, step
+from fab.steps import check_for_errors, run_mp, step
 from fab.util import TimerLogger, by_type
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


MatthewHambley (changes requested):
This looks like a useful change but your raising it pointed out that we didn't have a contributors agreement in place. We now do so you'll need to deal with that before this change can progress.

t00sa:
Thank you for your pull request.

In order to move forward with it, we require you to sign up to the [Momentum Open Source Software Contributor Agreement](https://github.com/MetOffice/Momentum/blob/main/CLA.md) and add your details to the table in the CONTRIBUTORS.md file to indicate your agreement.

If you do not wish to agree or are unable to do so, we will not be able to accept your contribution and will have to close the PR.

PR #580, opened 2026-06-25
Fail analyse step when parser errors occur

### LINKED ISSUE #577: Fab does not abort in case of analyse errors

jordansilly77-stack:
Closes #577

## Summary
- fail the analyse step when parser/analyser results contain exceptions
- reuse the existing error-reporting helper for consistent behaviour
- update unit coverage for the new failure path

## Tests
- `git diff --check`
- `uv run --with '.[dev]' python -m flake8 source/fab/steps/analyse.py tests/unit_tests/steps/test_analyse.py`\n- `uv run --with '.[dev]' python -m pytest tests/unit_tests/steps/test_analyse.py tests/unit_tests/steps/test_steps.py -q`

### DIFF
--- a/source/fab/steps/analyse.py
+++ b/source/fab/steps/analyse.py
@@ -46,7 +46,7 @@ from fab.mo import add_mo_commented_file_deps
 from fab.parse import AnalysedFile, EmptySourceFile
 from fab.parse.c import AnalysedC, CAnalyser
 from fab.parse.fortran import AnalysedFortran, FortranParserWorkaround, FortranAnalyser
-from fab.steps import run_mp, step
+from fab.steps import check_for_errors, run_mp, step
 from fab.util import TimerLogger, by_type
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


t00sa:
Thank you for your pull request.

In order to move forward with it, we require you to sign up to the [Momentum Open Source Software Contributor Agreement](https://github.com/MetOffice/Momentum/blob/main/CLA.md) and add your details to the table in the CONTRIBUTORS.md file to indicate your agreement.

If you do not wish to agree or are unable to do so, we will not be able to accept your contribution and will have to close the PR.

